PR #8, opened 2017-02-24
Bugfix/ls thing service breaks for bad codename

mkeshaw005:
Fixed an issue in ServerUtilityFunctions getFromACASServer where the json payload was being returned by the response object without being cast to string first.


mkeshaw005:
Just updated the request to merge with 1.10 instead of master.

PR #20, opened 2017-03-30
fix for running rapache in centos 7 environment with apache 2.4

brianbolt:


brianbolt:
going to fix on 1.10-release first and then merge to enhanced transactions

PR #25, opened 2017-04-05
Added new dynamic-loading Select2 ComboBox, a new LsThing route to ge…

bffrost:
…t codetables by label text, an internal version of the getNextLabelSequence route, and a new package.json dependency for parsing yaml. Also changed getFromACASServerInternal from returning JSON.stringify json to simply json.

### DIFF
--- a/modules/ServerAPI/src/server/routes/LabelServiceRoutes.coffee
+++ b/modules/ServerAPI/src/server/routes/LabelServiceRoutes.coffee
@@ -9,21 +9,25 @@ exports.getNextLabelSequence = (req, resp) ->
 		labelServiceTestJSON = require '../public/javascripts/spec/testFixtures/LabelServiceTestJSON.js'
 		resp.json labelServiceTestJSON.nextLabelSequenceResponse
 	else
-		config = require '../conf/compiled/conf.js'
-		baseurl = config.all.client.service.persistence.fullpath+"labelsequences/getLabels"
-		request = require 'request'
-		request(
-			method: 'POST'
-			url: baseurl
-			body: req.body
-			json: true
-		, (error, response, json) =>
-			if !error && response.statusCode == 200
-				resp.json json
-			else
-				console.log 'got ajax error trying to get next label sequence'
-				console.log error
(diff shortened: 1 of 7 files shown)


modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
```diff
@@ -51,11 +53,49 @@ exports.thingsByTypeKind = (req, resp) ->
...
+	request(
+		method: 'POST'
+		url: baseurl
+		qs: params
+		body: searchJSON
+		json: true
+	, (error, response, json) =>
+		console.log response.statusCode
+		if !error && response.statusCode == 200
+			callback json.results
```
brianbolt: We don't need to solve it here but no matter what, callback should be called. Otherwise clients don't get a response.

modules/ServerAPI/src/server/routes/LabelServiceRoutes.coffee
```diff
@@ -9,21 +9,25 @@ exports.getNextLabelSequence = (req, resp) ->
...
+		json: true
+	, (error, response, json) =>
+		if !error && response.statusCode == 200
+			callback json
+		else
+			console.log 'got ajax error trying to get next label sequence'
+			console.log error
+			console.log json
+			console.log response
+			callback "getNextLabelSequenceFailed"
```
brianbolt: this one, gets a call back no matter what. which is good. See other places where it doesn't

modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee
```diff
@@ -230,7 +230,7 @@ exports.getFromACASServerInternal = (baseurl, callback) ->
 		json: true
 	, (error, response, json) =>
 		if !error && response.statusCode == 200
-			callback response.statusCode, JSON.stringify json
```
bfielder: This change is going to break other modules. Need to revert it and add in an alternate route before merging this into anywhere

PR #37, opened 2017-04-21
check the logged in user, not the user in the sel file when evaluatin…

brianbolt:
…g the cross project loader roles

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -452,7 +452,7 @@ validateTreatmentGroupData <- function(treatmentGroupData,calculatedResults,temp
   }
   return(NULL) 
 }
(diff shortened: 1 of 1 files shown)


modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -577,12 +577,23 @@ validateCalculatedResults <- function(calculatedResults, dryRun, curveNames, tes
         rCompoundsDF <- batchProjectRestriced[batchProjectRestriced$isRestricted & batchProjectRestriced$Project.Code!=projectCode,]
         rCompounds <- rCompoundsDF$Requested.Name
         if (length(rCompounds) > 0) {
-          userRoles <- unlist(lapply(user$roles, function(role) role$roleEntry$roleName))
-          
-          if(configList$client.roles.crossProjectLoaderRole %in% userRoles) {
-            warnUser(paste0("Compounds from the following projects: ", paste(rCompoundsDF$name, collapse = "', '"),
-                            " will be loaded into project '",currentProj$name,"'"))
-          } else {
+          addProjectError <- TRUE
```
bffrost: Does this get set to FALSE if "shouldCheckRole" is false? I don't see an else block for the if(shouldCheckRole) block

Issue #12, opened 2017-03-01
CReg Edit Parent lets you save even if stereo category is unselected


bffrost:
Edit Parent workflow will show you the pink/red warning, but lets you save nonetheless.

Fortunately it doesn't persist the change to null out the stereo category - it remains what it was previously set to.

PR #46, opened 2017-04-28
Creg bug fix for updating parent with unselected stereo category

egao12:
Fixed bug to display error and not allow updating parent if stereo category is not selected


bffrost:
Fixes #12

PR #57, opened 2017-05-09
remove explicit jsonlite require or library

goshiro2010:


modules/ServerAPI/src/server/ImportCmpdRegProjectsIntoACAS.R
```diff
@@ -21,7 +19,7 @@ checkIfProjectExists <- function(projectName=""){
 	projectCurlURL <- paste0(racas::applicationSettings$client.service.persistence.fullpath,"lsthings/project/project?with=stub&labelType=name&labelKind=project name&labelText=")
 	fullProjectURL <- URLencode(paste0(projectCurlURL, projectName))
 	responseJSON <- getURL(fullProjectURL)
-	projects <- fromJSON(responseJSON)
```
brianbolt: i'm not 100% positive that the fromJSON here was originally the rjson package since jsonlite is being imported. I don't think r flips back and forth overriding each time you call library.  It might just do it once

PR #72, opened 2017-05-15
Feature/ls thing and form field jam

jmcneil86043:


modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -258,6 +259,136 @@ class window.ACASFormLSCodeValueFieldController extends ACASFormAbstractFieldCon
...
+	###
+		Launching controller must:
+		- Initialize the model with an LSInteraction
+    Do whatever else is required or optional in ACASFormAbstractFieldController
+	###
+	events: ->
+		"change select": "handleInputChanged"
+
+	template: _.template($("#ACASFormLSThingInteractionFieldView").html())
+	@
```
mkeshaw005: Seems like there shouldn't be an "@" sign hanging out here, or is this just weird github formatting artifact?

jmcneil86043: No, that is really in the code and shouldn't be there

mkeshaw005: Do you want to remove that before the branch is merged in?

modules/Components/src/client/PickList.coffee
```diff
@@ -544,15 +544,16 @@ class window.EditablePickListSelect2Controller extends EditablePickListSelectCon
...
 		@thingKind = @options.thingKind
 		@labelType = if @options.labelType? then @options.labelType else null
 		@placeholder = if @options.placeholder? then @options.placeholder else null
 		@queryUrl = if @options.queryUrl? then @options.queryUrl else null
+		unless @queryUrl? or (@thingType? and @thingKind?)
+			alert("ThingLabelComboBoxController URL misconfigured - crash to follow")
 
 	render: =>
-		@$el.select2
+		@selectController = @$el.select2
```
egao12: Do you reference @selectController when you instantiate a ThingLabelComboBoxController?

modules/Components/src/client/ACASFormFields.coffee
jmcneil86043: yes

modules/Components/src/client/PickList.coffee
jmcneil86043: No, I think I put that instance variable in then removed the code that referenced it. Should I remove that?

egao12: I think it's ok to leave it in there because you might need it for listening to event triggers. I was just wondering if there was an example where you used it for that.

PR #106, opened 2017-05-25
Feature/getcontainercodesbylabels v2 like maxresults

samanthamunoz:
Implemented getContainerCodesByLabelsLikeMaxResults which has the same function as getContainerCodesByLabels but with 2 additional parameters (like and maxResults)


samanthamunoz:
Yep! Will do that. @brianbolt

samanthamunoz:
Updated @brianbolt (removed comments)

PR #112, opened 2017-06-06
Feature/acas tag docker variable

brianbolt:


brianbolt:
Docker-compose and Docker will read a .env file in the directory they were executed and use the variables found in the file during builds, pulls...etc.  To avoid hardcoding tags in the repository, each of the tags defined in the docker-compose and docker files now use ${ACAS_TAG} to define the tag to pull down.  

example change:

```
-    image: mcneilco/acas-oss:RC-1.11.0
 +    image: mcneilco/acas-oss:${ACAS_TAG}
```


By default, the .env file is set to "latest".  To override this, you can:

1. export an environment variable in your shell

`export ACAS_TAG=develop`

2. change the .env file to point at your desired tags. **DO NOT CHECK THIS IN**

3. if you want to mix and match, you can still go to the docker-compose file and change the actual tag.  **DO NOT CHECK THIS IN**

PR #113, opened 2017-06-08
Create CurveCuratorRoutes.coffee

brianbolt:
added api routes for curve analysis

### DIFF
--- a/modules/CurveAnalysis/src/server/routes/CurveCuratorRoutes.coffee
+++ b/modules/CurveAnalysis/src/server/routes/CurveCuratorRoutes.coffee
@@ -1,4 +1,12 @@
 
+exports.setupAPIRoutes = (app, loginRoutes) ->
+	app.get '/api/curves/stubs/:exptCode', exports.getCurveStubs
+	app.get '/api/curve/detail/:id', exports.getCurveDetail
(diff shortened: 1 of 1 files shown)


goshiro2010:
I don't know if we need to expose all of the curve curator routes.
Maybe open up the following routes to view the curves

 +	app.get '/api/curves/stubs/:exptCode', exports.getCurveStubs
 +	app.get '/api/curve/detail/:id', exports.getCurveDetail
 +	app.get  '/api/curve/render/*', exports.renderCurve

brianbolt:
We almost always expose all api routes.  I think if we need to pick and choose routes then we would need to go through the whole set. I don't see any reason to start with this set of routes

Brian Bolt | John McNeil & Co. Inc. | 2223 Avenida de la Playa Suite 204, La Jolla, CA | 707.217.7598

goshiro2010:
3rd party app is requesting an api route to view the curves.
Should probably only open up the render route (if that is all that is required to view the curves):
app.get '/api/curve/render/*', exports.renderCurve

Don't want to expose put and post routes that modify data.

brianbolt:
Ok, I will remove those lines. Just reporting that this is unusual since 90% of routes have both api route and non/api route

PR #121, opened 2017-06-19
Service to get container information from batch code.

samanthamunoz:
service that takes in a batch code and gives all relevant container info

### DIFF
--- a/modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
+++ b/modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
@@ -53,6 +53,7 @@ exports.setupAPIRoutes = (app) ->
 	app.post '/api/createTubes', exports.createTubes
 	app.post '/api/throwInTrash', exports.throwInTrash
 	app.post '/api/updateContainerHistoryLogs', exports.updateContainerHistoryLogs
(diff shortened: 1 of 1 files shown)


modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
```diff
@@ -2304,3 +2305,43 @@ formatContainers = (containers, modifiedBy, modifiedDate, callback) ->
...
+	queryPayload =
+		{
+			"containerType": "well",
+			"containerKind": "default",
+			"stateType": "status",
+			"stateKind": "content",
+			"valueType": "codeValue",
+			"valueKind": "batch code",
+			"value": requestObject.batchCode,
+			"maxResults": 600,
```
brianbolt: I thought we were going to make max results a config

samanthamunoz: @brianbolt - where are configs pulled in from in base acas, same place as custom?

brianbolt:
This file is where all the default configs are set:
https://github.com/mcneilco/acas/blob/master/conf/config.properties.example

Client's can override these configs with their own config files.

samanthamunoz:
@brianbolt - this has been updated, ready for your review.

PR #122, opened 2017-06-19
Bugfix/clean rapache environment

brianbolt:


bffrost:
I tested this fix, and it works.

PR #131, opened 2017-06-23
Merge feature/abstractFormAndComponents_additionalServices into develop

egao12:
Merging to develop and then will make RC-1.12 off of it


modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -345,7 +357,7 @@ class window.ACASFormLSThingInteractionFieldController extends ACASFormAbstractF
 		else
 			opts.thingType = @thingType
 			opts.thingKind = @thingKind
-		@thingSelectController = new ThingLabelComboBoxController opts
+		@thingSelectController = new ThingLabelComboBoxController opt
```
jmcneil86043: This looks like the 's' on opts is dropped by accident

PR #134, opened 2017-06-27
Feature/websockets configuration

mkeshaw005:


conf/config.properties.example
```diff
@@ -513,3 +513,8 @@ client.projectEditor.restrictedCheckbox.show=true
 client.projectEditor.isRestricted.default=false
 
 client.roles.crossProjectLoaderRole=ROLE_ACAS-CROSS-PROJECT-LOADER
+
+#Add user roles that represent groups allowed to fast user switch
+#client.moduleMenus.fastUserSwitchingRoles=[]
+client.moduleMenus.fastUserSwitchingRoles=[{"lsKind":"PROJ-00000001","lsType":"Project","roleName":"User"},{"lsKind":"PROJ-00000002","lsType":"Project","roleName":"User"}]
```
brianbolt: Seems like the default should be empty here.  Existing ACAS instances would pick this up as the default when upgrading.

modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
```diff
@@ -25,7 +25,8 @@ exports.requiredScripts = [
 	'/lib/spin/js/spin.js',
 	'/lib/spin/js/jquery-spin.js',
 	'/lib/handsontable/dist/handsontable.full.js',
-	'/lib/select2-4.0.3/dist/js/select2.full.js'
+	'/lib/select2-4.0.3/dist/js/select2.full.js',
+    '/socket.io/socket.io.js'
```
brianbolt: an extra indent here

conf/config.properties.example
jmcneil86043: yes, I agree

PR #138, opened 2017-06-29
Fixes #137

### LINKED ISSUE #137: provide option to system integration tests to only run nondestructive tests
he current system tests include tests that change the database. This will allow you to run a subset that are read-only

server.systemTest.runDestructive=false

jmcneil86043:
you must add "-nondestructive" to a test you want to run when 
server.systemTest.runDestructive=false

### DIFF
--- a/modules/SystemTest/spec/serviceTests/SystemTest.coffee
+++ b/modules/SystemTest/spec/serviceTests/SystemTest.coffee
@@ -221,9 +221,6 @@ describe "D. Access to rApache -nondestructive", ->
 	describe "and then to racas runfunction", ->
 		before (done) ->
 			@.timeout(20000)
-			console.log "brian"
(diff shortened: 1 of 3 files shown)


modules/SystemTest/spec/serviceTests/SystemTest.coffee
```diff
@@ -221,6 +221,9 @@ describe "D. Access to rApache", ->
 	describe "and then to racas runfunction", ->
 		before (done) ->
 			@.timeout(20000)
+			console.log "brian"
+			console.dir JSON.stringify(runRFunctionServiceTestJSON.runRFunctionRequest)
+			console.log(config.all.client.service.rapache.fullpath + "runfunction")
```
brianbolt: Should remove these extra consoles

PR #142, opened 2017-07-10
CMG-535: Updated gulp script to generate sourcemaps for client code

koudns:


koudns:
Technically, a build distribution should be stand-alone and not depend on the existence of a parallel source directory tree.  This is the case for our acas Test (Stage, and Prod) environments; the parallel acas source directory tree is not made available in these environments.  So, if we compiled the .coffee from the source directory tree directing the output of the .js and .js.map into the build directory, as you would like it to work, the maps would point to the source directory tree which does not exist and the sourcemapping would fail.  I believe this is also the case for the Docker build environment where the acas base source directory is not normally made available to map to inside the Docker container.  

If we were to take a middle-ground approach and copy the .coffee into the build directory and then compile it there, we might encounter the problem where the .coffee is syntactically incorrect and fails to compile.  In this case, the .coffee and the .js and .js.map would be out of sync.  The old .js and .js.map would run but map to the new syntactically incorrect .coffee and that would be misleading.

So, I think don't think we can avoid the performance issue (A) at least the first time it's built.  The gulp tasks do run newer() so it shouldn't take too long when running subsequent builds.

For (B), I would just use .gitignore like you mentioned

Let me know if this makes makes sense and/or if you have other suggestions.

BTW, I'm not sure if if the branch I originally specified, feature-inventory, is the one that we should be merging into anymore.

PR #146, opened 2017-07-11
Feature/multiple labels acas form fields

bffrost:
Added new ACASFormField widget to control a list of labels. It is enabled by setting the "multiple":true attribute on formField configurations for a given label.


modules/Components/src/client/ACASFormMultiLabel.coffee
```diff
@@ -0,0 +1,89 @@
...
+		options =
+			modelKey: @options.labelKey
+			inputClass: @options.inputClass
+			formLabel: @options.formLabel
+			placeholder: @options.placeholder
+			required: @options.required
+			url: @options.url
+			thingRef: @options.thingRef
+			insertUnassigned: @options.insertUnassigned
+		@lc = new ACASFormLSLabelFieldController options
```
jmcneil86043: lc is too short and undescriptive for an instance variable name

PR #159, opened 2017-07-14
New service to get container logs from container codes.

samanthamunoz:
Issue: 158


modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
```diff
@@ -1855,6 +1856,36 @@ exports.getContainerLogsInternal = (labels, containerType, containerKind, labelT
...
+exports.getContainerLogsByContainerCodesInternal = (containerCodes, callback) =>
+	if global.specRunnerTestmode
+		inventoryServiceTestJSON = require '../public/javascripts/spec/ServerAPI/testFixtures/InventoryServiceTestJSON.js'
+		resp.json inventoryServiceTestJSON.getContainerAndDefinitionContainerByContainerLabelInternalResponse
+	else
+		console.debug "incoming getContainerLogsByContainerCodesInternal request: '#{JSON.stringify(containerCodes)}'"
+		exports.getContainersByCodeNamesInternal(containerCodes, (containersByCodeNamesResponse, statusCode) =>
+			response = []
+			for getContainer in containersByCodeNamesResponse
+				label = getContainer.container.lsLabels[0].labelText
```
brianbolt: what if the container has more than one label.  It seems like you should actually be filtering on preferred = true (should be only one preferred label)

PR #160, opened 2017-07-14
Feature/file form field

jmcneil86043:


modules/Components/src/client/ACASFormFieldsView.html
```diff
@@ -56,3 +56,11 @@
     </div>
 </script>
 
+<script type="text/template" id="ACASFormLSFileValueFieldView" xmlns="http://www.w3.org/1999/html">
+    <label class="control-label">launching controller must provide or set label</label>
+    <div class="controls">
+        <div class="bv_File span9" style="width:352px;"></div>
```
bffrost: should be bv_file

modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -510,3 +510,92 @@ class window.ACASFormLSDateValueFieldController extends ACASFormAbstractFieldCon
...
+		else
+			@$('.bv_File').html '<a href="'+window.conf.datafiles.downloadurl.prefix+fileValue+'">'+@getModel().get('comments')+'</a>'
+			@$('.bv_deleteSavedFile').show()
+
+	createNewFileChooser: ->
+		if @fileController?
+			@fileController.render()
+		else
+			@fileController = new LSFileChooserController
+				el: @$('.bv_File')
```
bffrost: should be bv_file

Issue #176, opened 2017-07-20
bugfix/Fix server side Thing, Container and experiment to create a new label after ignoring the old one


brianbolt:
New server utility function version of Label was ignoring the old value and trigger "createNewLabel" but the serverside Thing, Container and Experiment objects were not listening to the create

PR #177, opened 2017-07-20
#176: Listener to server side Thing, Experiment and Container objects

brianbolt:


modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee
```diff
@@ -836,10 +838,21 @@ class Thing extends Backbone.Model
 		if @lsProperties.defaultLabels?
 			for dLabel in @lsProperties.defaultLabels
 				newLabel = @get('lsLabels').getOrCreateLabelByTypeAndKind dLabel.type, dLabel.kind
+				@listenTo newLabel, 'createNewLabel', @createNewLabel
 				@set dLabel.key, newLabel
 				#			if newLabel.get('preferred') is undefined
 				newLabel.set preferred: dLabel.preferred
 
+	createNewLabel: (lKind, newText) =>
```
jmcneil86043: your trigger includes the key as the third argument, but you ignore it in this implementation. I think you should set the newLabel key. See Thing.coffee around line 222

modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee
```diff
@@ -1187,10 +1200,20 @@ class Container extends Backbone.Model
...
```
jmcneil86043: Is this function defined twice in this file?

jmcneil86043: Oh, part of a different class. Please set key here too

Issue #218, opened 2017-08-16
add option to moveToLocation to update the location history and make the default be false


brianbolt:
We need an options to update location history or not on moveToLocation.  The default for now should be false with an options to turn on by sending "1" to the updateLocationHistory parameter

PR #228, opened 2017-08-30
#218: add option to moveToLocation to optionally updateMoveToLocation…

brianbolt:
…, default for now will be false


modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
```diff
@@ -1366,22 +1369,27 @@ exports.moveToLocationInternal = (input, callCustom, callback) ->
 				if !error
 					console.log 'container down here...did it pull through??'
```
samanthamunoz: Can you remove this line? console.log 'container down here...did it pull through??'

PR #253, opened 2017-10-04
Feature/acas cmpd inv support

jmcneil86043:


modules/CmpdReg/src/client/custom/configuration.json
```diff
@@ -35,8 +35,9 @@
 	    "useExactMass": true,
 		"useProjectRolesToRestrictLotDetails": false,
 		"showTareWeight": false,
-		"showTotalAmountStored": false,
+		"showTotalAmoundStored": false,
```
bffrost: We don't need to reintroduce this spelling error. I have already fixed it both here and in the code that references it.

PR #271, opened 2017-10-11
Added line to hide Add Ls Role button after save comes back Fixes #270

### LINKED ISSUE #270: Label Sequence GUI should hide the Add LsRole button on save

bffrost:


bffrost:
Fixes #270

Issue #277, opened 2017-10-12
CReg - Click add alias clears selected option in type picklists


egao12:
When adding new aliases to a compound, clicking the Add button causes all the type dropdowns to clear

PR #278, opened 2017-10-12
bugfix/creg-fix-add-alias-clear-type-picklist-bug: Issue #277 (CReg -…

egao12:
… Click add alias clears selected option in type picklists) - need to update typeKind after type is selected


egao12:
Commit message should reference issue 277 not 227

Issue #286, opened 2017-10-17
Call Custom code was commented out of moveToLocation, it should be uncommented


brianbolt:
There are cases when we should call custom code when moveToLocation is complete, this code will allow this as an optional call

PR #287, opened 2017-10-17
#286: Uncommenting moveToLocation callCustom code

brianbolt:


modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
```diff
@@ -1371,15 +1371,15 @@ exports.moveToLocationInternal = (input, callCustom, updateLocationHistory, call
 					if updateLocationHistory
 						exports.updateContainerHistoryLogsInternal(input, (json, statusCode) ->
 							callback json, statusCode
```
samanthamunoz: Why is there a callback here before the "if"?

modules/ServerAPI/src/server/routes/InventoryServiceRoutes.coffee
```diff
@@ -1367,22 +1367,14 @@ exports.moveToLocationInternal = (input, callCustom, updateLocationHistory, call
...
+					callFunctionOrReturnNull shouldCallCustom, csUtilities.moveToLocation, input, (customerResponse, statusCode) ->
+						if updateLocationHistory
+							exports.updateContainerHistoryLogsInternal(input, (json, statusCode) ->
 								callback json, statusCode
-						)
-					else
-						callback json, response.statusCode
+							)
+						else
+							callback json, response.statusCode
```
samanthamunoz: I think if it gets to this else block you'll get an error right? json isn't defined.

brianbolt: why would I get an error?why would I get an error?

Issue #314, opened 2017-10-23
by default acas should require a few basic roles to be set


brianbolt:
ACAS should require users to have some basic roles by default.  We already have configs for ACAS-USER, ACAS-ADMIN, CMPDREG-ADMIN and more but they are not on by default

PR #315, opened 2017-10-23
#314: Turn on roles by default, configure module menu's to use basic …

brianbolt:
…default roles, give bob roles when running system test routes

### DIFF
--- a/conf/config.properties.example
+++ b/conf/config.properties.example
@@ -368,7 +368,7 @@ client.roles.acas.adminRole=ROLE_ACAS-ADMINS
 #client.roles.acas.userRole=
 #client.roles.acas.adminRole=
 
-#client.browser.enableSearchAll=true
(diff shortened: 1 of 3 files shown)


conf/config.properties.example
```diff
@@ -355,17 +355,20 @@ client.security.syncLdapAuthRoles=${server.security.syncLdapAuthRoles}
...
+client.roles.acas.adminRole=ROLE_ACAS-ADMINS
+
+# To remove roles set the roles configs to nothing.  You may also need to edit module menus configuration to not require these roles
+#client.roles.loginRole=
+#client.roles.cmpdreg.chemistRole=
+#client.roles.cmpdreg.adminRole=
+#client.roles.acas.userRole=
+#client.roles.acas.adminRole=
+
+#client.browser.enableSearchAll=true
```
bffrost: this line should not be commented out.

PR #323, opened 2017-10-24
Feature/updated merge and clone container services

bffrost:


gulpfile.coffee
```diff
@@ -43,10 +43,10 @@ getRPath = (path) ->
   return path
 
 getLegacyRPath = (path) ->
-  console.warn  "Warning: All R code in modules path 'modules/**/src/server/**/*.{R,r}' should be moved to 'modules/**/src/server/r/*.{R,r}'"
+  # console.warn  "Warning: All R code in modules path 'modules/**/src/server/**/*.{R,r}' should be moved to 'modules/**/src/server/r/*.{R,r}'"
```
bffrost: I don't think these warnings should be commented out. The point of them is to spur people to move files in modules in the legacy path format into the updated path format. If all the R files are put in the right place, we won't see these warnings on compile.

brianbolt:
I agree, but I need to do it better since it was giving the wrong path.  I will add it back in correctly later.


Brian Bolt | John McNeil & Co. Inc.  | 707.217.7598

bffrost:
this commit is not needed for the upcoming 1.13 release, and will be merged in at a later point

Issue #328, opened 2017-10-25
add 'Aliquot' to configured entity types and preferred entity services


brianbolt:
'Aliquot' is a container tube with a physical state of 'solution'

PR #331, opened 2017-10-25
#328: add 'Aliquot' to configured entity types and preferred entity s…

brianbolt:
…ervices

### DIFF
--- a/conf/ConfiguredEntityTypes.coffee
+++ b/conf/ConfiguredEntityTypes.coffee
@@ -74,3 +74,12 @@ exports.entityTypes =
     sourceExternal: false
     parent: false
     model: require("../routes/ServerUtilityFunctions.js").DefinitionContainerTube
+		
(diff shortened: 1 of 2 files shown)


conf/ConfiguredEntityTypes.coffee
```diff
@@ -1,76 +1,85 @@
...
+    code: 'Definition Container Tube'
     type: 'definition container'
     kind: 'tube'
     codeOrigin: 'ACAS LsContainer'
     displayName: 'Definition Tube'
     sourceExternal: false
     parent: false
     model: require("../routes/ServerUtilityFunctions.js").DefinitionContainerTube
+  ,
+    code: 'Aliquot'
```
bffrost: I think the code "Aliquot" could be something more clear - like "Solution Container Tube" to specify that it is explicitly solution phase. it would need to be changed below as well.

PR #352, opened 2017-10-29
Fixes #351 numeric form field rounding interfered with input

### LINKED ISSUE #351: Bug where numeric form field rounding caused inout interference
The fixed digit option was updating the field while the user was typing

jmcneil86043:


bffrost:
[Amount field should force 3 digits on change of focus, not on keyup](https://trello.com/c/wo0RjEY7/167-amount-field-should-force-3-digits-on-change-of-focus-not-on-keyup)

PR #356, opened 2017-10-30
Fixes #355 where location tree filter input has error styling when fo…

### LINKED ISSUE #355: When the location form field shows error style, the tree filter inout is also red
The error styling that applies to the form field is also applied to the tree inouts when opened and should not be

jmcneil86043:
…rm field shows error


bffrost:
[When making a new lot aliquot entry, the location filter input is marked as required](https://trello.com/c/qIHaeuS9/175-when-making-a-new-lot-aliquot-entry-the-location-filter-input-is-marked-as-required)

PR #369, opened 2017-11-03
Changed PreferredEntityCodeService.getSpecificEntityType to be more f…

### LINKED ISSUE #350: SEL should give a more helpful error message when there is an issue with the entity type given in the SEL file
This is the current error message: "Corporate Batch ID is not valid in the first column. It should be something like 'Corporate Batch ID'."

Would be nice to have a list of accepted entity types.

### LINKED ISSUE #368: When loading an SEL file with a solution aliquot id-specified in the protocol, SEL rejects the experiment and gives conflicting errors in the GUI

bffrost:
…orgiving and lookup by code if the lookup by displayName fails.

Fixed reporting of invalid entityType to list available configured entityTypes. Fixes #350
Switched getPreferredId2 call to use displayName instead of code. Added function to dealias entity displayName or code into code. Switched entity type not matching that required by the protocol to be an error, not a warning. Fixed reporting of entity ype mismatch to use displayNames, not codes. Fixes #368


bffrost:
@brianbolt Please give this one a thorough review

Issue #374, opened 2017-11-08
Create node route to move to location with a location breadcrumb input


samanthamunoz:
Perform the move to location using a location breadcrumb as the input in lieu of a location container code.

PR #375, opened 2017-11-08
Issue #374 Feature/node route move to location by breadcrumb

samanthamunoz:


samanthamunoz:
@bffrost - all fixed!

PR #408, opened 2017-11-17
fixes #407: Add mechanism for having a configured entity type that is…

### LINKED ISSUE #407: Add mechanism for having a configured entity type that is not loadable as a tested entity
Currently there is no way to add a configured entity type that is not allowed to be loaded as a tested entity.  This becomes an issue when you have entity types like "Solution Container Tube" that check projects of their content "Corporate Batch ID" thus requiring a configured entity type for their contents, however we don't want to allow users to load that entity type.

brianbolt:
… not loadable as a tested entity

Currently there is no way to add a configured entity type that is not allowed to be loaded as a tested entity. This becomes an issue when you have entity types like "Solution Container Tube" that check projects of their content "Corporate Batch ID" thus requiring a configured entity type for their contents, however we don't want to allow users to load that entity type.

### DIFF
--- a/modules/ServerAPI/src/server/routes/PreferredEntityCodeService.coffee
+++ b/modules/ServerAPI/src/server/routes/PreferredEntityCodeService.coffee
@@ -1,6 +1,8 @@
 exports.setupAPIRoutes = (app) ->
 	app.get '/api/entitymeta/configuredEntityTypes/:asCodes?', exports.getConfiguredEntityTypesRoute
 	app.get '/api/entitymeta/configuredEntityTypes/displayName/:displayName', exports.getSpecificEntityTypeRoute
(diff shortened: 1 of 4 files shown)


modules/ServerAPI/src/server/routes/PreferredEntityCodeService.coffee
```diff
@@ -57,6 +61,38 @@ exports.getSpecificEntityType = (codeNameOrDisplayName, callback) ->
...
+	if asCodes
+		codes = for type in testedEntityTypes
+			code: type.code
+			name: type.displayName
+			ignored: false
+		return codes
+	else
+		return testedEntityTypes
+
+exports.getSpecificLoableEntityTypeRoute = (req, resp) ->
```
bffrost: What is this function name? Seems like it should be exports.getSpecificTestedEntityTypeRoute

PR #432, opened 2017-12-15
Bugfix/audit config defaults rc 1.13.0

bffrost:


modules/CmpdReg/src/client/custom/configuration.json
```diff
@@ -1,112 +1,131 @@
...
+		"connectToServer": true,
+		"baseServerURL": "/cmpdreg/",
+		"logoutURL": "/logout",
+		"acasURL": "http://localhost:8080/acas/api/v1/",
+		"acasAppURL": "http://localhost:3001/api/"		 
+	},
+	"clientUILabels": {
+		"corpNameLabel": "CMPD Number",
+		"applicationNameForTitleBar": "ACAS Compound Registration",
+		"commonNameLabel": "Original ID:"
```
jmcneil86043: I think "Common Name" would be better for default

modules/CmpdReg/src/client/custom/configuration.json
```diff
@@ -1,112 +1,131 @@
...
+		"corpSeparator": "-",
+		"saltSeparator": "",	
+		"batchSeparator": "-",
+		"numberCorpDigits": 7,
+		"formatBatchDigits": 3,
+		"startingCorpNumber": 0,
+		"fancyCorpNumberFormat": false,
+		"corpParentFormat": "standard",
+		"corpBatchFormat": "corp_batch_saltcode",
+		"appendSaltCodeToLotName": true,
```
jmcneil86043: let's not show the salt code in the lot name -> false

modules/CmpdReg/src/client/custom/configuration.json
bffrost: The actual UI field that uses this config was taken out when we added Aliases, so it is a dead config

jmcneil86043: ok then

modules/CmpdReg/src/client/custom/configuration.json
```diff
@@ -1,112 +1,135 @@
...
+				{"name": "Lot Salt Equivalents", "dataType": "string", "required": false},
+				{"name": "Lot Solution Amount Units", "dataType": "string", "required": false},
+				{"name": "Lot Solution Amount", "dataType": "numeric", "required": false},
+				{"name": "Lot Stock Location", "dataType": "string", "required": false},
+				{"name": "Lot Stock Solvent", "dataType": "string", "required": false},
+				{"name": "Lot Storage Location", "dataType": "string", "required": false},				
+				{"name": "Lot Supplier ID", "dataType": "string", "required": false},
+				{"name": "Lot Supplier Lot", "dataType": "string", "required": false},
+				{"name": "Lot Supplier", "dataType": "string", "required": false},
+				{"name": "Lot Synthesis Date", "dataType": "date", "required": false},
```
fkmeyer: @bffrost don't we almost always have the date marked as required? Seems like it wouldn't be a bad default setting.

modules/CmpdReg/src/client/templates/AppControllerView.inc
```diff
@@ -31,7 +31,7 @@
 
 	</div>
 
-	<div id="footer"><p>&#169; 2011-2015 John McNeil & Co., Inc.</p></div>
+	<div id="footer"><p>&#169; 2011-2017 John McNeil & Co., Inc.</p></div>
```
fkmeyer: Should we go ahead and bump to 2018? @jmcneil86043 @bffrost

modules/ModuleMenus/src/client/ModuleMenusConfiguration.coffee
```diff
@@ -31,12 +27,6 @@ window.ModuleMenusConfiguration =
...
 		requireUserRoles: [window.conf.roles.acas.userRole]
-	,
-		isHeader: false
-		menuName: "Project Editor"
-		mainControllerClassName: "ProjectController"
-		autoLaunchName:"project"
-		requireUserRoles: [window.conf.roles.acas.adminRole]
 	,
 		isHeader: false, menuName: "Example Thing"
 		mainControllerClassName: "ExampleThingController"
```
fkmeyer: Do we need to hide Example Thing?

modules/CmpdReg/src/client/custom/configuration.json
bffrost: We have had requests in the past to have the required set be only the set actually required by the database, so the most minimal set. This is why Lot Synthesis Date got changed (a couple versions ago) from required to not in the default bulk loader configs.

modules/ModuleMenus/src/client/ModuleMenusConfiguration.coffee
bffrost: @jmcneil86043 ?

modules/CmpdReg/src/client/templates/AppControllerView.inc
bffrost: I defer to John on this one

modules/CmpdReg/src/client/custom/configuration.json
fkmeyer: Ah okay cool. I was thinking about it in the context of the single reg gui anyway. Makes sense!

PR #457, opened 2018-01-30
fixes #456 : Fix "Use Existing Experiment" as a way to delete and rel…

### LINKED ISSUE #456: Fix "Use Existing Experiment" as a way to delete and reload data against experiment code
Currently "Uses Existing Experiment" does not work and does not have enough error reporting.

brianbolt:
…oad data against experiment code

Currently "Uses Existing Experiment" does not work and does not have enough error reporting.


brianbolt:
@egao12

Issue #467, opened 2018-02-13
Add config for whether SOLR index should be updated when things are saved/updated


egao12:
Add config for whether solr should be updated on post/put of thing. Update SOLR route will be a customer specific function.

PR #468, opened 2018-02-13
feature/update-solr-for-things-config: Fixes issue #467. Added config…

egao12:
… to update solr when things are saved/updated, refactored thing service routes to update solr if config is set, trigger model change when tags are changed.


egao12:
Will merge into release/1.13 later

Issue #469, opened 2018-02-23
ld_entitlements.py call fails with Error: spawn E2BIG if json argument string is too long


bffrost:
This indicates node is hitting an underlying OS limit for command-line calls, MAX_ARG_STRLEN
Suggested solution: Convert json string with data contents to a temp file and call ld_entitlements.py with that file as input instead of specifying the json as a command-line argument.

PR #476, opened 2018-03-05
bugfix/sync-project-users-write-groups-json-to-temp-file: fixes Issue…

egao12:
… #469 - wrote groupsJSON to temp file, which will be read by ld_entitlements.py

### DIFF
--- a/modules/ServerAPI/src/server/python/syncProjectsUsers/ld_entitlements.py
+++ b/modules/ServerAPI/src/server/python/syncProjectsUsers/ld_entitlements.py
@@ -20,6 +20,10 @@ API = '/api/'
 
 
 def parse_options(argv):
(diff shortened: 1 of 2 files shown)


modules/ServerAPI/src/server/python/syncProjectsUsers/ld_entitlements.py
```diff
@@ -20,6 +20,10 @@
 
 
 def parse_options(argv):
+    groupsJSONFile = open(argv[1], "r")
```
bffrost: I think this will leave the file open for reading. I don't think it matters deeply, but it is better form to do something like this, which will close the file after the "with" block ends
with open(argv[1], "r") as groupsJSONFile:
    import_json = groupsJSONFile.read()
    argv[1] = import_json

Issue #519, opened 2018-07-24
Cleanup wrongly checked in files


brianbolt:
There should be no files in the public/javascripts there are currently a number of files that are being picked up by the client that need to be deleted.

PR #522, opened 2018-07-26
Added column for "Restricted Data" to Project Browser summary table v…

### LINKED ISSUE #349: Project Browser should show Is Restricted in table view, either as a new column or instead of Project Leaders
The Is Restricted value is a much more important attribute than Project Leaders.

bffrost:
…iew. Fixes #349


bffrost:
I think this bug is caused by #519
I'll merge release/1.13.2 into this feature branch and re-test.

brianbolt:
It's possible this was never an issue or it's possible that pulling in release/1.13.2 fixed the issue but either way it's not an issue now after testing with the current branch.

bffrost:
I'm still having issues with the current branch in dev mode. Let's hold off on merging this until I get these resolved.

PR #527, opened 2018-08-09
fixes #526: Allow manual assignment of lot number in manual registrat…

### LINKED ISSUE #526: Allow manual assignment of lot number in manual registration for new lots
Introduce a new set of options for metaLot in configuration.json:

Require that lot numbers are filled in before registration:
```
		"requireLotNumber": true,
```

Allow users to manual enter lot numbers:
```
		"allowManualLotNumber": true,
```

Auto populate the next lot number into the lot number field during registration
```
		"autoPopulateNextLotNumber": true
```

If requireLotNumber is true, then either allowManualLotNumber or autoPopulateNextLotNumber is required to be true.

If any of these options are true, the code should fetch the current set of lot numbers.  If autopopulate is set to true, the code should add one to the max of the lot numbers and supply it to the lot number field.  If allow manual lot number is set to true, then the user can edit the lot number field but the code will present an error along with a list of already used lot numbers if the user has entered an already existing lot number.

brianbolt:
…ion for new lots

### DIFF
--- a/modules/CmpdReg/src/client/custom/Lot_Custom.js
+++ b/modules/CmpdReg/src/client/custom/Lot_Custom.js
@@ -35,7 +35,10 @@ $(function() {
       totalAmountStored: null,
       totalAmountStoredUnits: null,
       vendor: null,
-      vendorID: null
+      vendorID: null,
+      parent: null,
+      lotNumbers: null
+      // maxAutoLotNumber: 1000
     },
     
     initialize: function(){
@@ -57,14 +60,14 @@ $(function() {
                     vendor: new PickList(js.vendor)
                 })
         // replace composite object pointers with real objects
-        this.set({fileList: new BackboneFileList()});
+        this.set({fileList: new BackboneFileList()}, {silent: true});
         this.get('fileList').model = BackboneFileDesc;
         this.get('fileList').add(js.fileList);
 
       } else if (this.isNew()){
         this.set({
           fileList: new BackboneFileList()
-        });
+        }, {silent: true});
         this.get('fileList').model = BackboneFileDesc;
       }
     },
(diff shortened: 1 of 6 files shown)


modules/CmpdReg/src/client/custom/Lot_Custom.js
```diff
@@ -160,19 +163,76 @@ $(function() {
...
+              errors.push({'attribute': 'lotNumber', 'message':  "Lot Number must be an integer"});
+            } else {
+              if(attr.lotNumber == 0) {
+                errors.push({'attribute': 'lotNumber', 'message':  "Lot Number 0 is reserved for virtual lots"});
+              } else {
+                lotNumbers = this.get('lotNumbers');
+                if(lotNumbers.includes(attr.lotNumber)) {
+                  errors.push({'attribute': 'lotNumber', 'message':  "This lot number is already taken by one of the lots for this compound "+lotNumbers.join(',')});
+                }
+                //  else {
```
bffrost: Can this section of commented-out code be removed?

modules/CmpdReg/src/client/custom/configuration.json
```diff
@@ -39,7 +39,10 @@
 		"showTotalAmountStored": false,
 		"disableAliasEdit": false,
 		"showLotInventory": false,
-		"disableEditMyLots": false
+		"disableEditMyLots": false,
+		"requireLotNumber": true,
+		"allowManualLotNumber": true,
```
bffrost: So the new default behavior will be to allow a manual lot number, but auto-populate with the next one? Sounds reasonable to me, but it is a default behavior change.

brianbolt: Thanks for pointing this out. I checked in the wrong default configs.  I want the default behavior to not change.

modules/CmpdReg/src/client/custom/Lot_Custom.js
brianbolt: I think you are referring to the commented out part. I actually should just add some comments around it.  It's an example of how one might allow the manual assignment of lot numbers greater than some max number configuration, but anything less than this number must be auto assigned.

brianbolt:
I forgot that the configuration.json is not like config.properties.example.  I'll add some code in to check the configuration and assume default behavior if they are not set.

Issue #544, opened 2018-12-20
ACAS falsely reports that experiment already exists in another protocol


brianbolt:
ACAS reports that an experiment which was previously loaded exists in a different protocol from that of the upload file. 

```
Experiment 'X' does not exist in the protocol that you entered, but it does exist in 'Y'. Reloading the file will update the data and change the protocol.
```

PR #549, opened 2019-01-05
Fix for recently introduced SEL bug when re-loading existing experiment

bffrost:
Fix for bug introduced in commit 78d4d413462a5c2ce5ceb27bc6410d4f0dc7c4d9 where "id" was being compared to "codeName" to determine whether a protocol matched, rather than comparing "codeName" to "codeName".

Resulting bug was the wrong warning when a user tried to reload an experiment to the same protocol -- warning that the experiment exists on another protocol, rather than just warning that the data will be deleted.


brianbolt:
This was filed with #544 and fixed with https://github.com/mcneilco/acas/releases/tag/1.13.3.6 (haven't merged that back to master which is why you hit it)

PR #553, opened 2019-01-17
Feature/acas cmpdreg merge

brianbolt:


brianbolt:
This is probably this biggest change I have made to acas in the last 2 years.  I would appreciate input from both @jmcneil86043 and @bffrost when you have time.  I am happy to walk through some of the decisions I made and answer any questions you have.  The idea here is that that compound reg roo server is now merged to acas roo server.  Additionally the migration involves removing the scientist, altering column types of creg and merging the two schemas which requires a manual script to be run as a super user (which I can point you to).  

There are other things blocking this from being deployed so this is not a high priority but i'd like to know if there are major concerns in the next 2 weeks or so so  I can get them straightened out.

modules/CmpdReg/src/client/src/Parent.js
```diff
@@ -32,7 +32,7 @@ $(function() {
 					exactMass: js.exactMass,
                     molFormula: js.molFormula,
 					comment: js.comment,
-                    chemist: new PickList(js.chemist),
+                    chemist: js.chemist,
```
bffrost: I'm confused about how a PickList with the current chemist selected is getting replaced with just the raw chemist code. Could you clarify @brianbolt ?

modules/CmpdReg/src/client/src/SearchForm.js
```diff
@@ -7,7 +7,7 @@ $(function() {
             _.each(attributes, function(att, key) {
                 if( (att != '' && att!=null) && key!='aliasContSelect' && key!='searchType' && key!='percentSimilarity' && key!='maxResults' && key!='loggedInUser' ) {
                     if( key=='chemist') {
-                        if( att.get('code')!='anyone') {
```
bffrost: Is there a corresponding place where this "anyone" chemist is defined that needs to be updated to go with the new scheme of chemist-as-string instead of chemist-as-object?

modules/CmpdRegBulkLoader/src/server/routes/CmpdRegBulkLoaderRoutes.coffee
```diff
@@ -166,28 +196,38 @@ exports.registerCmpds = (req, resp) ->
...
-						console.log json
-						console.log response
-						if json.summary?
-							resp.json [json]
+				exports.getScientistsInternal (authors) =>
+					_ = require 'underscore'
+
+					authorCodes = _.pluck authors, "code"
+
+					exports.validationPropertiesInternal req.body, (sdfProperties) =>
```
bffrost: It's great to add this upfront validation of the Scientists referenced in the file. What do you think about doing the same thing for Stereo Categories and Projects?

modules/Login/src/server/routes/loginRoutes.coffee
```diff
@@ -222,4 +222,14 @@ exports.getAuthors = (req, resp) ->
 		baseEntityServiceTestJSON = require '../src/javascripts/spec/testFixtures/BaseEntityServiceTestJSON.js'
 		resp.end JSON.stringify baseEntityServiceTestJSON.authorsList
 	else
-		csUtilities.getAuthors req, resp
+		if csUtilities.getAuthors?
```
bffrost: Is the point of this to preserve pre-existing CustomerSpecificServerFunctions at clients where we tie into an external authentication service? Just making sure I understand.

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
```diff
@@ -0,0 +1,189 @@
+"""
```
bffrost: I'm unclear on where, if it all, this entire python file is being referenced/used. Am I missing something, or has that yet to be added?

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
```diff
@@ -0,0 +1,189 @@
...
+    auth_return = client.auth_check(username = username, password = password)
+    return auth_return
+
+def ld_group_to_acas_role(group):
+    # u'LDAP_Auto_ROLE_SEURAT-USERS'
+    # u'Project_ProjectX_Administrator'
+    split = group["name"].split('_')
+    ls_type = split[0]
+    ls_kind = split[1]
+    role_name = "_".join(split[2:len(split)])
```
bffrost: We've discussed this elsewhere, so I'm OK with it as part of this merge, but I think we should get away from doing string manipulation like this to map roles if possible.

modules/ServerAPI/src/server/routes/ExperimentServiceRoutes.coffee
```diff
@@ -490,11 +490,15 @@ exports.genericExperimentSearch = (req, res) ->
...
-		baseurl = config.all.client.service.persistence.fullpath+"experiments/search?q="+req.params.searchTerm+"&userName="+username
-		console.log "baseurl"
-		console.log baseurl
-		serverUtilityFunctions = require './ServerUtilityFunctions.js'
-		serverUtilityFunctions.getFromACASServer(baseurl, res)
+		authorRoutes = require './AuthorRoutes.js'
+		authorRoutes.allowedProjectsInternal req.user, (statusCode, allowedUserProjects) ->
+			_ = require "underscore"
+			allowedProjectCodes = _.pluck(allowedUserProjects, "code")
+			baseurl = config.all.client.service.persistence.fullpath+"experiments/search?q="+req.params.searchTerm+"&projects=#{allowedProjectCodes.join(',')}"
```
bffrost: Does this same refactor need to be done in the ProjectBrowser module's search method? That is the other place where we limit results back based on accessible projects.

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
```diff
@@ -0,0 +1,189 @@
...
+        'id': ld_user["id"],
+        'username': ld_user["username"],
+        'email': ld_user["username"],
+        'firstName': ld_user["username"],
+        'lastName': ld_user["username"],
+        'roles': roles
+    }
+    return acas_user
+
+def get_users_roles(client, users):
```
bffrost: Nit: This might make more sense if you defined a python class that had "client" as an instance attribute (i.e. "self.client") rather than having to pass it into many functions as an argument. Just food for thought, not a required change.

modules/CmpdRegBulkLoader/src/server/routes/CmpdRegBulkLoaderRoutes.coffee
brianbolt: Will resolve with #554

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
brianbolt: This will be used in CustomerSpecificServerFunctions.coffee

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
brianbolt: Agreed

modules/ServerAPI/src/server/routes/ExperimentServiceRoutes.coffee
brianbolt: Yes it should and this will be resolved with #555

modules/ServerAPI/src/server/python/acas_ldclient/acasldclient.py
brianbolt: I will likely need to make more changes to this client so I will make this change then

modules/CmpdReg/src/client/src/Parent.js
brianbolt: The picklist, which was really just the chemist object was required previously as it contained the id and other attributes required by ROO to match the chemist back to the scientist table.  We are now just passing the string along.

PR #559, opened 2019-02-11
fixes #558: Additional codes in scientist picklists

### LINKED ISSUE #558: Additional codes in scientist picklists
We need to add the ability for non-authors or registered users to show up in picklists in ACAS.  The use case of this is that Compounds are often synthesized and Assays performed by an external user or CRO which may not be a registered user. Validation and picklists should combine the current list of registered users with a data dictionary list for 2 different types of purposes:

1) Picklists and validation for experiment, protocol modules should come from a dictionary with:

```
codeType: assay
codeKind: scientist
```

2) Picklists and validation for compound registration should come from a dictionary with:

```
codeType: chemist
codeKind: scientist
```

brianbolt:


modules/ServerAPI/conf/ProtocolConfJSON.coffee
```diff
@@ -470,11 +470,11 @@
...
 				ignored: false
 			,
-				codeType: "model fit"
-				codeKind: "transformation units"
+				codeType: "compound"
+				codeKind: "scientist"
 				codeOrigin: "ACAS DDICT"
-				code: "%"
-				name: "%"
+				code: "Pharmaron"
```
bffrost: Is it intentional that this unrelated ddict value was replaced by an example compound scientist?

brianbolt: Yes, this was a mistake, fixing now.

Issue #564, opened 2019-03-05
Update README for ACAS to reflect that Dockerfile-indigo is required for OSS builds


brianbolt:

simonkeng:
Hi @brianbolt, 

Related to updating the README, I noticed that `jchem-merged-lib-16.4.25.0.zip` file mentioned in the [Compound Registration ROO Server setup](https://github.com/mcneilco/acas#add-chemaxon-jar-file) section of the README is not available any longer from [ChemAxon's website](https://chemaxon.com/download?dl=%2Fdata%2Fdownload%2Finstant-jchem%2F16.4.25.0).

I'm wondering which would be the equivalent, perhaps `instantjchem-16.4.25.0.zip`?

brianbolt:
Took a little digging in around their website but you can actually still download jchem 16 here:

https://chemaxon.com/download?dl=%2Fdata%2Fdownload%2Fjchem%2F16.4.25.0

Here is  a direct link to the referenced merged zip file:

https://chemaxon.com/download?dl=%2Fdata%2Fdownload%2Fjchem%2F16.4.25.0%2Fjchem-merged-lib-16.4.25.0.zip

We do plan on updating the instructions for newer jchem versions but those require a different type of build process since the jars are split.  We either need to point at chemaxons maven repo using an API key or a hosted internal repo that has all of the jars.

simonkeng:
That's great, I knew it had to be somewhere. Thanks for digging it up!

PR #566, opened 2019-05-17
Update README, provide mock user creation script

simonkeng:
### Updated README:

- Provide updated link to zip containing `jchem.jar`, provided by @brianbolt on #564 
- Provide mock (bob) user creation instructions. 
- Provide instructions for bringing the stack down. 
- Added shell script for creating initial user for local development environment, script provided by @bffrost


simonkeng:
Hi @bffrost, that makes sense. I wasn't sure which branch would be best for the PR. Will do. Thanks.

PR #569, opened 2019-07-02
Feature/compound assay scientist editors

brianbolt:
I removed the CmpdReg Scientist module and added 2 new modules for editing the code tables for Compound Scientist (codeType: compound, codeKind: scientist) and Assay Scientist (codeType: assay, codeKind: scientist).


modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminView.html
```diff
@@ -0,0 +1,125 @@
+<script type="text/template" id="AbstractCodeTablesAdminView" xmlns="http://www.w3.org/1999/html">
+    <div class="form-horizontal">
+        <div class="control-group">
+            <label class="control-label bv_group_codeTablesAdminCode"><%=upperDisplayName%> Code</label>
```
bffrost: Should this have an asterisk too? When I played with the UI it seemed Code and Name were both required, but only Name had the * next to it

brianbolt: Thanks, yes, I made it required, i'll add the asterisk

PR #571, opened 2019-07-08
fixes #570

### LINKED ISSUE #570: Picklists should be limited by application system role
If `client.roles.acas.userRole` or `client.roles.cmpdreg.chemistRole` are defined, they should be used to limit the picklists in acas and cmpdreg

brianbolt:


brianbolt:
> Changes LGTM. Assuming there is a second PR coming for acas-roo-server with the new codetable route for getting by authorRole only?

I already merged the acas-roo-server change because it was simple. 

https://github.com/mcneilco/acas-roo-server/commit/70c7a71c9f0626346710e95ba2ca6a5fa7e500f1

bffrost:
Those changes look good too!

PR #593, opened 2019-10-03
fixes #592 add validation mode to compound reg bulkloader

### LINKED ISSUE #592: Bulk compound registration validation mode

brianbolt:


brianbolt:
@bffrost Thanks for the review, I will take a look at the issues you highlighted and resolve them accordingly.



> if the new validationMode.enable config is turned on, users will HAVE to validate first, then will be given the option to proceed to registration depending on whether the validation finds errors and that allowRegistrationOnError config is true. Similar to SEL, makes sense.

Yes, the goal was to make it like just SEL if the right settings were configured.

PR #605, opened 2019-11-21
Improved audit table handling for dateValue and codeValue

bffrost:
Make it so both table and edit value view show human-readable values for codeValues and dateValues

### DIFF
--- a/modules/Components/src/client/UtilityFunctions.coffee
+++ b/modules/Components/src/client/UtilityFunctions.coffee
@@ -86,3 +86,4 @@ class window.UtilityFunctions
 		else
 			console.log "can't find entry in pickLists hash for: "+value.get('lsKind')
 			console.dir pickLists
+			return "not found"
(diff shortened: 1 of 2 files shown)


modules/Components/src/client/UtilityFunctions.coffee
```diff
@@ -77,3 +77,12 @@ class window.UtilityFunctions
...
 			return Math.round((num+0.000001)*1000)/1000
+	
+	getNameForCode: (value, codeToLookup, pickLists) ->
+		if pickLists[value.get('lsKind')]?
+			code = pickLists[value.get('lsKind')].findWhere({code: codeToLookup})
+			name = if code? then code.get 'name' else "not found"
+			return name
+		else
+			console.log "can't find entry in pickLists hash for: "+value.get('lsKind')
+			console.dir pickLists
```
brianbolt: It's not clear what gets returned here if there is no match.

brianbolt:
Made one comment but otherwise LGTM. Good to merge now or you can look into my comment if you think it would ever be an issue.

PR #641, opened 2020-01-22
fixes #640 check for label uniqueness and regex formatting string

### LINKED ISSUE #640: Thing label options for regex validation and uniqueness checking

brianbolt:


modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -183,24 +183,73 @@ class window.ACASFormLSLabelFieldController extends ACASFormAbstractFieldControl
...
+				if value != ""
+					@getModel().set
+						labelText: value
+						ignored: false
+				else
+					@setEmptyValue()
+			callback()
+
+	checkUnique: (value, callback) ->
+		if @getModel().get("validationRegex")?
```
bffrost: I get the logical flow, but it's strange to me that the "validationRegex" check is under a function called "checkUnique".

modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -183,24 +183,73 @@ class window.ACASFormLSLabelFieldController extends ACASFormAbstractFieldControl
...
+			# so we don't need to do that here
+			if(value == "") 
+				callback(true)
+				return
+			# check for uniqueness
+			$.ajax
+				type: 'POST'
+				url: "/api/getThingCodeByLabel/#{@getModel().get("uniqueBy").thingType}/#{@getModel().get("uniqueBy").thingKind}"
+				data: 
+					thingType: @getModel().get("uniqueBy").thingType
```
bffrost: Is there a reason this is passed in via "uniqueBy" rather than just directly setting the thingType and thingKind on the Label model? If we're setting new properties anyway, I find "uniqueBy" kindof misleading. I feel like I should need to pass in "labelText" or some other parameter to tell ACAS which field(s) to "unique by"

modules/Components/src/client/ACASFormFields.coffee
brianbolt: Yeah definitely not clear and easy to read or easily reusable.  I just pushed an update which I think is much better.

PR #662, opened 2020-03-10
fixes #661 Provide bootstrap main function

### LINKED ISSUE #661: Bootstrap script should provide main and callback
Currently the bootstrap script is run on startup and runs without a callback.  There are scenarios in which you would want to detect when the bootstrap script has completed by providing a callback function.  

- The Boostrap should provide a main function with a callback return for an error string
- The ACAS template should be backwards compatible in case no main function is provided

brianbolt:


modules/ServerAPI/src/server/Bootstrap.coffee
```diff
@@ -22,3 +26,5 @@ if config.all.server.liveDesign.installClientOnStart? && config.all.server.liveD
 #     if error?
 #         console.warn "Error installing ld client. This can happen if Live Design was down during ACAS start or otherwise ACAS could not communicate with Live Design.  This is just a warning as LD client might already be installed and be the correction version."
 #         console.error stderr
+if require.main == module
```
bffrost: Sorry I don't understand this syntax at all. What does this do?

brianbolt: This is the nodejs equivalent\of python's if __name__ == '__main__'

so if you call it as a script like `node Bootstrap.js` it will just run the main function
but if you require it in code, you can import the functions and call them individually as the change to the app_api_template.coffee file does as part of this change.

bffrost: Got it.

PR #669, opened 2020-03-19
667 adding route for getting curve metadata quickly in bulk by curve id

brianbolt:


brianbolt:
Whoops accidently pre-merged this one

bffrost:
I did look at this but my R is not strong enough to really review, sorry.

brianbolt:
NBD, its largely based on pre-existing code, except for the query and it all works so should be fine.  I might start pulling in @philippcheung to look at R changes if he's willing.

PR #672, opened 2020-04-10
send author roles instead of username to the updated POST service

brianbolt:


bffrost:
@brianbolt is there an accompanying change in Roo?

brianbolt:
Yep, thought I requested you on it, just added you.  Here is the link:

https://github.com/mcneilco/acas-roo-server/pull/173

PR #683, opened 2020-04-24
fixes #682 add route for bulk posting things and moving their files to the correct place based on controller redirect

### LINKED ISSUE #682: Add route for bulk posting ls things and moving files to configured redirect conf ls thing file path

brianbolt:


modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
```diff
@@ -285,6 +287,42 @@ updateThing = (thing, testMode, callback) ->
...
+			filesToSave = fileVals.length
+
+			completeThingUpdate = (thingToUpdate)->
+				updateThing thingToUpdate, false, (updatedThing) ->
+					callback updatedThing, 200
+
+			fileSaveCompleted = (passed) ->
+				if !passed
+					callback "file move failed", 500
+				if --filesToSave == 0 then completeThingUpdate(thing)
```
bffrost, line 303: What is this `--filesToSave` doing? Decrementing by one?

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
```diff
@@ -285,6 +287,42 @@ updateThing = (thing, testMode, callback) ->
...
+
+			if filesToSave > 0
+				prefix = serverUtilityFunctions.getPrefixFromEntityCode thing.codeName
+				for fv in fileVals
+					csUtilities.relocateEntityFile fv, prefix, thing.codeName, fileSaveCompleted
+			else
+				callback thing, 200
+
+		if response.indexOf("saveFailed") > -1
```
bffrost, line 312: oh, this is where the actual function calls for the callback from `bulkPostThingsInternal` start.

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
bffrost: I'm really confused about where the actual changes to the fileValues happen. We seem to just be passing the `thing` that comes back from the "bulk post" and then calling "updateThing" on it. When do the file values actually change?

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
```diff
@@ -285,6 +287,42 @@ updateThing = (thing, testMode, callback) ->
...
+
+			fileSaveCompleted = (passed) ->
+				if !passed
+					callback "file move failed", 500
+				if --filesToSave == 0 then completeThingUpdate(thing)
+
+			if filesToSave > 0
+				prefix = serverUtilityFunctions.getPrefixFromEntityCode thing.codeName
+				for fv in fileVals
+					csUtilities.relocateEntityFile fv, prefix, thing.codeName, fileSaveCompleted
```
bffrost, line 308: Is `fv` being modified by this function, and the change flowing all the way back to `thing`?

brianbolt: relocate entity file is modifying the thing and then calling its call back "fileSaveCompleted".  This way, the customer specific server function is responsible for doing something with the file.

brianbolt:
I'll add comments to the code. Most of this was copied from function within anothing thing save route.

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
brianbolt: I added comments to clarify but yes, it is decrementing by one.

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
bffrost: I guess I am just confused that this `relocateEntityFile` function is not having the `thing` variable passed to it. How does it access the `thing`? Does it look it up by codeName?

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
bffrost: Similar question here... This `fileSaveCompleted` function is only getting the true/false `passed` variable as an input. Where is the value of `thing` coming from?

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
brianbolt: its getting passed just the file value `fv`. I just added some more code comments to clarify (I had to look it up to).

PR #693, opened 2020-05-06
Fix for authorized label sequences

brianbolt:


modules/ACASLabelSequences/src/server/routes/ACASLabelSequencesRoutes.coffee
```diff
@@ -50,7 +50,7 @@ exports.getAuthorizedLabelSequencesInternal = (req, callback) ->
 		callback labelSequenceTestJSON.labelSequenceArray
 	else
 		roles = req.session.passport.user.roles
-		baseurl = config.all.client.service.persistence.fullpath+"labelsequences/getAuthorizedLabelSequences"
```
bffrost, line 53: This looks like it will produce the url:
`labelsequences/getAuthorizedLabelSequences?&thingTypeAndKind=...`
Is the `?&` instead of just `?` going to be a problem?

brianbolt: No it is not a problem.  It's not the cleanist but its valid and I tested it and it works on all 4 cases (without query params, with labelTypeAndKind, with thingTypeAndKind and with both of those query params)

PR #709, opened 2020-06-19
fixes #708 add POST request option for rendering curves

### LINKED ISSUE #708: curve urls are too long
URLs can only be  ~2000 characters.  We should allow a POST request for a curve render

brianbolt:


modules/CurveAnalysis/src/server/renderCurve.R
```diff
@@ -3,16 +3,25 @@
...
   # Redirect to Curator if applicable
   redirectInfo <- racas::api_get_curve_curator_url(getParams$curveIds, getParams$inTable, globalConnect = TRUE)
   if(redirectInfo$shouldRedirect == TRUE) {
     setHeader("Location", redirectInfo$url)
     return(HTTP_MOVED_TEMPORARILY)
     DONE
   }
+
+  postParams <- NA
+  saveSession('/tmp/ham')
```
bffrost, line 18: Is this on purpose? Looks like a debug line

PR #726, opened 2021-01-07
fixes #725 cmpd reg button cutoff

### LINKED ISSUE #725: Cmdreg metalot form buttons cut off on bottom of page
Buttons are still just barely clickable but user has to remember which is which.

Screen Shot 2021-01-07 at 10 35 40 AM

brianbolt:
This fixes a long issue seen on Safari which made its way to Chrome as chrome changed the way it handles something with our css/html

Screen Shot 2021-01-07 at 10 44 51 AM


bffrost (approved):
LGTM

PR #738, opened 2021-03-12
fixes #737 Allow lot project service to return multiple rows per lot

### LINKED ISSUE #737: Get preferred ID should be allowed return multiple projects for a lot
Currently the lot project service is required to return one and only one project for a lot.  If a customer uses a system which allows lots to be associated with multiple projects and also uses project restrictions, then experiment loader won't work correctly for them.  The behavior is that experiment loader will report that the lot belongs to another restircted project even if one of the lots projects matches the experiment project.
 
Experiment loader should allow the service to return both projects and if one of them matches the experiment project, then they can proceed.

brianbolt:


bffrost (approved):
This is WAY clearer. I can follow the code now and it makes sense to me. Thank you!

PR #742, opened 2021-05-08
741 acas forms blob value support

brianbolt:


modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
```diff
@@ -63,10 +65,55 @@ exports.setupRoutes = (app, loginRoutes) ->
...
+			resp.status(200).send();
+
+exports.bufferToStream = (buffer) ->
+	Duplex = require('stream').Duplex
+	stream = new Duplex
+	stream.push buffer
+	stream.push null
+	stream
+
+exports.getlsValuesInternal = (id, params, callback) ->
```
bffrost, line 109: @brianbolt Is this a duplicate of `exports.getlsValuesByIdInternal` up above or is there something subtly different that I'm missing?

modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -785,6 +789,82 @@ class ACASFormLSFileValueFieldController extends ACASFormAbstractFieldController
...
+		else
+			displayText = @getModel().get('comments')
+			if !displayText?
+				displayText = "InternalErrorUnknownFileName"
+			id = @getModel().get('id')
+			# Display inline not supported right now because guessing mimetype from extension isn't available
+			# so always be falsey here but leaving this in place
+			if false and @displayInline and @mimeType?
+				@$('.bv_file').html '<img src="data:' + @mimeType + ';base64,'+@arrayToBase64String(fileValue)+'" alt="'+ displayText+'">'
+				@$('.bv_file').html '<img src="/api/thingvalues/downloadThingBlobValueByID/'+id+'" alt="'+ displayText+'">'
```
bffrost, line 824: Why are we setting the html to two different values on two consecutive lines? This seems like we're missing another if/else?

modules/ServerAPI/src/server/routes/FileServices.coffee
```diff
@@ -21,13 +23,61 @@ setupRoutes = (app, loginRoutes, requireLogin) ->
...
+							"size": file.size,
+							"type": file.mimetype,
+							"deleteType": "DELETE",
+							"url": "/tempfiles/" + path.basename(file.path),
+							"binaryData": binaryData
+						}
+						files.push(outfile)
+					resp.json {"files": files}
+				catch err
+					console.error("AAAAAND")
```
bffrost, line 67: nit: please remove debug logging

bffrost (commented):
Can't say I fully understand the multer stuff but this generally makes sense to me. Added a few minor comments but otherwise looks good

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
brianbolt: This one is defining the function and the one above is calling.

modules/ServerAPI/src/server/routes/FileServices.coffee
brianbolt: addressed in a new commit

modules/Components/src/client/ACASFormFields.coffee
brianbolt: addressed in a new commit by removing the second line (this whole block is not used as noted above but its good to remove so later it won't cause confusion)

brianbolt:
> Can't say I fully understand the multer stuff but this generally makes sense to me. Added a few minor comments but otherwise looks good

The multer funciton is just a request middleware that takes multipart posts of files and writes them to a directory so that they are immediately available in the request.  By using it only in the blob route it won't affect other routes.  However, at somepoint in the future we should replace the jquery-file-upload-middleware package with multer which is the express recommended middleware (atleast according to their upgrade guide)

PR #744, opened 2021-05-08
fixes #743 buble duplicate compound error to user with status code 409

### LINKED ISSUE #743: Duplicate compound error not reported to user
Certain configurations don't allow status code 500 from node server to bubble up to the user.  Change status code to 409 when duplicate compound triggered from single reg.

brianbolt:


bffrost (approved):
LGTM

Issue #745, opened 2021-05-13
Picklists are not limiting by ls type and ls kind


brianbolt:
Request uses wrong names "thingType" instead of "lsType" and "thingKind" instead of "lsKind" [acas/ThingServiceRoutes.coffee at 27589b44f5269e6b332c6d6af047a4dd70c1cd90 · mcneilco/acas · GitHub](https://github.com/mcneilco/acas/blob/27589b44f5269e6b332c6d6af047a4dd70c1cd90/modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee#L97)

LsThingQueryDTO used in this route: [acas-roo-server/LsThingQueryDTO.java at 422fb9b9ac81622727596b4058914696db21da74 · mcneilco/acas-roo-server · GitHub](https://github.com/mcneilco/acas-roo-server/blob/422fb9b9ac81622727596b4058914696db21da74/src/main/java/com/labsynch/labseer/dto/LsThingQueryDTO.java#L23)

PR #749, opened 2021-05-22
#745 Render blob values as base 64 downloads when value is not saved

brianbolt:


bffrost (approved):
LGTM. If I'm understanding right this is for unsaved forms after the initial upload of a blob but before LsThing save.

brianbolt:
Yes, part of the basic thing validate and save component.

PR #753, opened 2021-05-22
fixes #752 acas thing browser

### LINKED ISSUE #752: Add a generic thing browser to acas base

brianbolt:


modules/ServerAPI/src/client/Thing.coffee
```diff
@@ -187,6 +189,17 @@ class Thing extends Backbone.Model
 			if !data.has ('recordedDate') || data.get('recordedDate') == null
 				data.set recordedDate: rDate
 
+	setName: =>
```
brianbolt, line 192: This name is not saved in the database but rather is used for UI components (such as the new ACASThingBrowser) to be able to provide a default display name for the thing type/kind.

modules/ServerAPI/src/client/Thing.coffee
```diff
@@ -187,6 +189,17 @@ class Thing extends Backbone.Model
...
+		# This is a best guess as a pretty name using the ls type and kind
+		# it can be overriden by setting the name of the ls thing when
+		# instantiating the ls thing
+		if !@name?
+			typeName = @.get("lsType").charAt(0).toUpperCase() + @.get("lsType").slice(1).toLowerCase();
+			kindName = @.get("lsKind").charAt(0).toUpperCase() + @.get("lsKind").slice(1).toLowerCase();
+			@name = "#{kindName} #{typeName}"
```
brianbolt, line 199: I found that `"#{kindName} #{typeName}"` was typically a better looking name than "#{typeName} #{kindName}" but this can always be overridden by passing the name when instatiating the new thing (or by setting it in the thing conf)

modules/ServerAPI/src/client/Thing.coffee
```diff
@@ -437,7 +456,7 @@ class Thing extends Backbone.Model
 			version: 0
 
 	getStateValueHistory: (vKind) =>
-		valInfo = _.where(@lsProperties.defaultValues, {key: vKind})[0]
+		valInfo = @getValueInfo(vKind)
```
brianbolt, line 459: Pulled this into a function.

modules/ServerAPI/src/server/routes/ControllerRedirectRoutes.coffee
```diff
@@ -62,11 +62,18 @@ exports.redirectToEditor = (req, resp) ->
...
+			if controllerRedirectConf[queryPrefix]?
+				if controllerRedirectConf[queryPrefix][kind]?
+					deepLink = controllerRedirectConf[queryPrefix][kind]["deepLink"]
+					resp.redirect "/"+deepLink+"/codeName/"+code
+				else
+					resp.statusCode = 500
+					resp.end "Could not find #{queryPrefix}:#{kind} in controller redirect configuration"				
+			else 
+				resp.statusCode = 500
+				resp.end "Could not find #{queryPrefix} in controller redirect configuration"
```
brianbolt, line 76: The new ACASThingBrowser needs a way to automatically determine the deep link.  We actually have this setup in various places by configuring the ControllerRedirectConf.  This change is just to add better messages when the controller redirect conf is misconfigured.

modules/ServerAPI/src/client/Thing.coffee
bffrost: I think `name` is too vague for this. Maybe "Class Name" or "Thing Kind Display Name"? I originally thought this meant a preferred label, and I could see this being confusing as many Things have "names" that are particular to an instance rather than the name of the whole class of Things.

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -0,0 +1,313 @@
...
+		@$(".bv_thingController").removeClass("hide")
+		@$(".bv_thingControllerContainer").removeClass("hide")
+
+		@$('.bv_editThing').show()
+		if window.conf.thing?.editingRoles?
+			editingRoles = window.conf.thing.editingRoles.split(",")
+			if !UtilityFunctions::testUserHasRole(window.AppLaunchParams.loginUser, editingRoles)
+				@$('.bv_editThing').hide()
+
+        # Not allowing deleting Thing right now
```
bffrost, line 257: Don't understand this comment as I do see logic around deleting Things below.

bffrost (approved):
LGTM, but would prefer a rename of the `name` attribute to something more specific to prevent confusion in the future.

bffrost (approved):

modules/Components/src/client/ACASThingBrowser.coffee
brianbolt: The code below was commented out when I started due to an issue but I fixed it and now I'm removing the comment about deleting being disabled.

PR #758, opened 2021-05-28
757 allow filter by labelType and labelKind for getThingCodeByLabel

brianbolt:
Body was always taking precedence over what was passed as a parameter in the URL so I made it optional to pass in the parameters in the URL and left it so that body continues to override the URL parameters.  Throws a 400 if thingType or thingKind are not passed in either.


bffrost (approved):
LGTM. Didn't know about that `qs` argument but that definitely helps make this cleaner.

PR #760, opened 2021-06-02
Don't error when calling to json on thing which has a missing state

### LINKED ISSUE #759: Don't error when calling to json on thing which has a missing state
Thin.toJSON errors if a state is missing from the Thing which is in the defaultProperties. It should ignore this issue.

brianbolt:
Fixes #759


bffrost (approved):
LGTM

PR #761, opened 2021-06-03
752 acas thing browser class

brianbolt:
This fixes a few issues with ACASThingBrowser:

- TinyMCE wasn't rendering when ThingController was instatiated so Clob values were not shown
- Adds the ability to search by recorded date with searches like
  - "2021" all recorded date within year 2021
  - "2021-04 all recorded date within April 2021
  - "2021-04-02 all recorded date on April 2nd 2021
- Adds the ability to search by recordedBy
- Refactors ACASThingBrowser to not require a separate query config
  - To update to this branch, classes that extend this class need to
    - Rename "toDisplay" option to "configs"
    - Remove "query" option
  - To display a thing attribute without making it searchable you can add `false` to the `isSearchable` key of the config you are adding (default is to always search by all display values.

### DIFF
--- a/modules/Components/src/client/ACASThingBrowser.coffee
+++ b/modules/Components/src/client/ACASThingBrowser.coffee
@@ -239,12 +239,14 @@ class ACASThingBrowserController extends Backbone.View
 			$(".bv_thingTableController").html @thingSummaryTable.render().el
 
 	selectedThingUpdated: (thing) =>
+		@$('.bv_thingControllerWrapper').append("<div class='bv_thingController'></div>")
 		@trigger "selectedThingUpdated"
 		@thingController = new @controllerClass
+			el: @$('.bv_thingController')
 			model: thing
 			readOnly: true
 
-		@$('.bv_thingController').html @thingController.render().el
+		@thingController.render()
 		@$(".bv_thingController").removeClass("hide")
 		@$(".bv_thingControllerContainer").removeClass("hide")
 
--- a/modules/Components/src/client/ACASThingBrowser.coffee
+++ b/modules/Components/src/client/ACASThingBrowser.coffee
@@ -88,36 +88,39 @@ class ThingSimpleSearchController extends AbstractFormController
 
 	getQueryTerms: (queryTerms, searchTerm) ->
 		for queryValue in @configs
-			#Default is all display values are searchable
-			isSearchable = !queryValue.isSearchable? || queryValue.isSearchable != false
-			# Code Name is part of the default search so skip it here
-			if isSearchable && queryValue.key != "codeName"
-				if queryValue.key == "recordedDate"
-					@addRecordedDateToQuery(queryTerms, searchTerm)
+			# Code Name and Recorded By are part of the defaults set them as isSearchable false
+			if ["codeName", "recordedBy"].includes(queryValue.key)
+				queryValue.isSearchable = false
+	
+			# Default is all display values are searchable so if the attribute is missing or set to
+			# anything other than false, then it is searchable
+			isSearchable = (!queryValue.isSearchable? || queryValue.isSearchable != false)
+			if queryValue.key == "recordedDate"
+				@addRecordedDateToQuery(queryTerms, searchTerm)
+			else
+				#If the key is an ls value
+				valDef = @model.getValueInfo(queryValue.key)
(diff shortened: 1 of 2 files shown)


modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -111,27 +86,92 @@ class ThingSimpleSearchController extends AbstractFormController
 					@$(".bv_thingSearchTerm").attr "disabled", false
 					@$(".bv_doSearch").attr "disabled", false
 
+	getQueryTerms: (queryTerms, searchTerm) ->
+		for queryValue in @configs
+			#Default is all display values are searchable
+			isSearchable = !queryValue.isSearchable? || queryValue.isSearchable != false
+			# Code Name is part of the default search so skip it here
+			if isSearchable && queryValue.key != "codeName"
```
bffrost, line 94: Don't quite follow why we need to filter out `codeName`, but if so, wouldn't we also want to filter out `recordedBy`?
Seems like a more generic check for whether a particular key is in the defaults would be preferable. I feel like I may be missing something here though.

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -111,27 +86,92 @@ class ThingSimpleSearchController extends AbstractFormController
...
+								operator: searchOperator
+		return queryTerms
+
+	addRecordedDateToQuery: (queryTerms, searchTerm) ->
+		# Default search for recordedDate is ISO without time
+		# e.g. 2021-06-02
+		dateParts = searchTerm .split('-')
+		# Offset the user entered date by one month to account for month = 0 in javascript
+		if typeof(dateParts[1]) != "undefined"
+			dateParts[1] = dateParts[1]-1
```
bffrost, line 129: Little worried here if I type in dash-separated terms that are not dates. Will this bomb out if I type in `brian-bolt-rules` ?

bffrost (changes requested):
Overall makes sense to me. I'm in favor of making this change now before we've built a lot on top of this. Commented on two areas which didn't quite make sense to me.

modules/Components/src/client/ACASThingBrowser.coffee
brianbolt: Yeah, understandable and so was I.  Essentially most cases a user enters will not return a valid date and therefore won't be added to the query because there is a check on the date (`!isNaN(recordedDateGreaterThan))`)

Some examples:
`1234-1234-1234`: Pared into, `Mon Feb 15 1340 16:07:02 GMT-0752` and added to the query (no results because its a crazy date)
`brian-bolt-rules`: Invalid date, not added to query
`2021-1-Monday`: Invalid date, not added to query

In other parts of code we do test for a "reasonable date" but I did not add that here becaues unreasonable dates are unlikely to have real matching values so I didn't think it would hurt much.

modules/Components/src/client/ACASThingBrowser.coffee
brianbolt: Your right about this.  Its just that the way I coded it, codeName is always searched on so there is nothing to do here (its already part of the queryParams).  RecordedBy is not always searched on (only if its in the display and isn't marked as isSearchable: false).  I'll clean this up to make it clear.  I realize now with the recordedBy  that there may be fields where you want to search on it, but not display it which I should as a feature.

modules/Components/src/client/ACASThingBrowser.coffee
bffrost: I think this stems from my lack of understanding of Javascript. As long as line 131:
```recordedDateGreaterThan = new Date(Date.UTC(...dateParts))```
just produces `recordedDateGreaterThan = undefined` or NaN, then the code makes sense to me as-is. Was just confused that we cast to date without any try/catch.

modules/Components/src/client/ACASThingBrowser.coffee
brianbolt: Retracting my last statement. Recorded By is always searched on so I just refactored the code to exclude those.  I could check the keys of queryTerms to check for their existence instead of the hardcoded way I wrote it.

bffrost (approved):
LGTM

PR #767, opened 2021-06-08
Basic validate and save is broken due to file chooser api change, adding .gitignore items

### LINKED ISSUE #766: Basic validate and save is broken due to file chooser api change
File chooser returns a file object rather than just a name now.  Controllers that use the file chooser controller API need to use the returned file objects `file.name` instead of expecting just a string with the filename.

brianbolt:
Fixes #766


bffrost (approved):
LGTM

PR #776, opened 2021-07-02
New rows in a state table are not rendering automatically so defaults are not rendered properly and can show as an error when they are not.

### LINKED ISSUE #774: New rows in a state table are not rendering automatically so defaults are not displayed
New rows in state tables are not rendering automatically when being added so if a value has defaults, they are not being displayed.  The symptom in this case is that required fields with default values are shown as errors.

brianbolt:
Fixes #774


bffrost (approved):
LGTM

PR #777, opened 2021-07-02
State table rows with code values which are null should be shown as empty when not required and null values

### LINKED ISSUE #775: State table rows with code values which are null should be shown as empty when not required and null values
Currently state rows with null values are shown with a red "not found" and in error even if the field is not required.

brianbolt:
Fixes #775


bffrost (approved):
LGTM

PR #780, opened 2021-07-13
Fixing redirect to client.baseBath

brianbolt:
Small bug fix to fix redirects to "/".  We should respect the config `client.basePath`


bffrost (approved):
LGTM

PR #787, opened 2021-07-15
ACASFormStateTableController should be able to start with 0 empty rows if configured

### LINKED ISSUE #784: ACASFormStateTableController should be able to start with 0 empty rows if configured
There are cases where we don't want to display any rows in an ACASFormStateTableController if no values are present.  In this case we should be able to add a configuration for startRows: 0 and minRows:0 to show an empty table.  This ticket is to implement the startRows feature as minRows already exists.

brianbolt:
Fixes #784


bffrost (approved):
LGTM

PR #790, opened 2021-07-28
Save original file name in comments when saving ls through ls file chooser

### LINKED ISSUE #789: Save original file name in comments when saving ls through ls file chooser

brianbolt:
Fixes #789


brianbolt:
I verified this worked on both ls blob and ls file values.  In both cases the original file name was maintained.

bffrost (approved):
LGTM

PR #792, opened 2021-08-18
ACAS-133: Add simple textarea clobValue FormField

bffrost:
HTMLClobValue was causing problems because it includes HTML in the saved value, and we were trying to use it to store command-line arguments which should be plaintext.
Implemented an "ACASFormLsClobValueController" tied to type `clobValue` (as opposed to `htmlClobValue`) which creates a simple `textarea` which is resizable.


brianbolt (changes requested):

brianbolt (approved):
LGTM

PR #795, opened 2021-09-03
Issue793

brianbolt:


modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
```diff
@@ -78,73 +76,29 @@ getRFileHandlerString = (rFilesWithRoute, config, acasHome)->
...
-					if option?
-						option = option[0].replace(' -D ','')
-						option = option.split('=')
-						option = {option: option[0], value: option[1]}
-						compileOptions.push(option)
-
-		console.log apacheVersion
-		compileOptions.push(option: 'ApacheVersion', value: apacheVersion)
-		compileOptions
+	compileOptions = [ { option: 'ApacheVersion', value: 'Redhat' },
```
bffrost, line 79: Am I understanding right that the other OS options have been removed because containerized is now the only supported deployment pattern?

modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
```diff
@@ -236,96 +161,76 @@ getApacheSpecificConfString = (config, apacheCompileOptions, acasHome) ->
...
-	else
-		apacheSpecificConfString = ''
+	apacheSpecificConfString = getApacheSpecificConfString(config, apacheCompileOptions, acasHome)
 	rapacheConfString = getRApacheSpecificConfString(config, apacheCompileOptions, acasHome)
 	rFilesWithRoute = getRFilesWithRoute()
 	rFileHandlerString = getRFileHandlerString(rFilesWithRoute, config, acasHome)
 	fs.writeFileSync "#{acasHome}/conf/compiled/apache.conf", [apacheSpecificConfString,rapacheConfString,rFileHandlerString].join('\n')
 	fs.writeFileSync "#{acasHome}/conf/compiled/rapache.conf", [rapacheConfString,rFileHandlerString].join('\n')
 
-csUtilities.getConfServiceVars sysEnv, (confVars) ->
```
bffrost, line 249: I guess we are removing support for an external config service here, which was likely only used at one now-defunct customer.

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
+# Process configs
```
bffrost, line 1: It would be helpful somewhere to document what this script is for. As I understand it is not directly used during startup - it is just a utility for admins to convert their legacy config files to environment variables, right?
Some explanation of this should either be added to the top of this script, to the README, or both.

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+    return(out_lines, name_maps)
+
+def process_configs(args):
+    out_lines = []
+    name_maps = {}
+    [example_lines, example_name_maps] = process_property_lines(open("conf/config.properties.example").readlines())
+    if args.properties:
+        lines = args.properties.readlines()
+
+        # lines = get_text_file_lines("conf/config.properties.example")
```
bffrost, line 81: some commented out code from testing that can be removed.

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+        [out_lines, name_maps] = process_property_lines(lines, list(example_name_maps.keys()))
+
+    if args.cmpdreg:
+        # out_lines.append("\n# CmpReg settings\n")
```
bffrost, line 85: Same here

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+            env_name = key.replace('.','_').upper()
+            setting = cmpdreg_configs[key]
+            if isinstance (setting, bool):
+                setting = json.dumps(setting)
+
+            name_maps[key] = {
+                            "env_name": env_name,
+                            "setting": str(setting)+"\n"
+            }
+            # out_line = f"{key}=${{env.{env_name}}}\n"
```
bffrost, line 99: And here

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+            # out_lines.append(out_line)
+
+
+    env_file_lines = []
+    env_file_lines.append("ACAS_TAG=release-1.13.6\n\n")
```
bffrost, line 104: This seems like it should be a variable, right?

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+    with open(file_path, 'r') as f:
+        lines = f.readlines()
+    return lines
+
+def write_lines_to_file(file_path, lines):
+    with open(file_path, 'w') as f:
+        for line in lines:
+            f.write(line)
+
+def process_property_lines(lines, example_names=None):
```
bffrost, line 26: Could you add a docstring explaining what `example_names` does? I can sortof get it (only looking for novel configs) but I'm still confused.

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+                out_line = ""
+            else:
+                out_line = line
+        out_lines.append(out_line)
+    return(out_lines, name_maps)
+
+def process_configs(args):
+    out_lines = []
+    name_maps = {}
+    [example_lines, example_name_maps] = process_property_lines(open("conf/config.properties.example").readlines())
```
bffrost, line 77: Where is this script supposed to be run from that this filepath works?

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
```diff
@@ -0,0 +1,146 @@
...
+
+    env_file_name = "new-environment.env"
+    write_lines_to_file(env_file_name, env_file_lines)
+    print(f"Wrote environment variables to {env_file_name}")
+    properties_file_name = "new-config.properties"
+    write_lines_to_file(properties_file_name, out_lines)
+    print(f"Wrote properties to {properties_file_name}")
+
+
+argparser = argparse.ArgumentParser(description='Process configs')
```
bffrost, line 133: Could use a more verbose explanation of what this script produces. It seems to produce two files: `new-config.properties` which lists only novel config files not found in `config.properties.example` and a `new-environment.env` which lists only override values which don't match the default values.

But I'm confused about how / where it's supposed to be invoked.

.env
```diff
@@ -1 +1,917 @@
-ACAS_TAG=latest
+ACAS_TAG=release-1.13.6
```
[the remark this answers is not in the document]
brianbolt: I did always mean to flip this back but meant to do it before the pull request. Going to revert it to latest

modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
brianbolt: Thats right. I don't think any of these systems would work anymore, or atleast their configuration would be difficult. This was unused code.

modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
brianbolt: Yeah. This is not used any longer.  I don't think its a use case we need any more so this was unused code and was actually breaking running prepare config files when customer specific server functions isn't available like in the context of racas and roo. I want to maintain the same script between all 3.

modules/BuildUtilities/src/server/python/zzz-config-to-envs.py
brianbolt: This script needs a lot of work. I might spin out another ticket for that instead of resolving all the issues as part of this merge. I will actually remove it here.

bffrost: Fair enough. I'm fine with you leaving it in as long as there's another ticket spun out for cleaning it up. I just picture it prompting a lot of questions if you point an SA at it right now.

conf/docker/roo/environment/roo.env
```diff
@@ -1,6 +0,0 @@
-HIBERNATE_DIALECT=org.hibernate.dialect.PostgreSQLDialect
-DATABASE_DRIVER=org.postgresql.Driver
-DATABASE_URL=jdbc:postgresql://db:5432/${ACAS_DB_NAME}
-VALIDATION_QUERY=select version()
-CATALINA_OPTS="-Xms512M -Xmx1536M -XX:MaxPermSize=512m"
```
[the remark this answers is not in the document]
brianbolt: These are the replacement environment variables in ACAS:

SERVER_DATABASE_DRIVER
SERVER_DATABASE_URL
SERVER_SERVICE_PERSISTENCE_JDBC_VALIDATION_QUERY


CATALINA_OPTS was always optional. Its also set in the docker-compose files and k8s so no need for it here.

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
```diff
@@ -128,6 +127,7 @@ exports.getBasicCmpdReg = (req, resp) ->
 	console.log 'in getBasicCmpdReg'
 	console.log req.originalUrl
 	endOfUrl = (req.originalUrl).replace /\/cmpdreg\//, ""
+	console.log "Config #{config.all.client.service.cmpdReg.persistence.basepath}"
```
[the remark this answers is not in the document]
brianbolt: I removed this

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
```diff
@@ -18,7 +18,7 @@ exports.setupRoutes = (app, loginRoutes) ->
 	app.get '/cmpdReg/parentAnnotations', loginRoutes.ensureAuthenticated, exports.getBasicCmpdReg
 	app.get '/cmpdReg/fileTypes', loginRoutes.ensureAuthenticated, exports.getBasicCmpdReg
 	app.get '/cmpdReg/projects', loginRoutes.ensureAuthenticated, exports.getAuthorizedCmpdRegProjects
-	app.get '/cmpdReg/vendors', loginRoutes.ensureAuthenticated, exports.getBasicCmpdReg
```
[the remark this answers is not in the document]
brianbolt: Thanks. I switched it when I hit a 500 error on that route and then it worked.  I didn't end up digging into it until I saw your comment.  I was missing an @Autowired in roo.  I just fixed roo and returned this back to the original line to avoid any issues.

bffrost (approved):
LGTM. Reviewed the latest set of changes and they all look good.
Similar to the other PR, would you mind going through my comments and resolving them once you believe they've been fixed?

PR #799, opened 2021-09-12
Ls thing code value picklists should have an option to add a new code value

brianbolt:
Fixes #798

PR #802, opened 2021-09-16
Revert "Ls thing code value picklists should have an option to add a new code value"

brianbolt:
Reverts mcneilco/acas#799


brianbolt:
@anand-schrodinger @bffrost Sorry for the issues. Reverting this for now and I'll fix and add it back later.

bffrost:
Thanks @brianbolt . I'd be happy to review when you add it back in. Admin-editable codeValue picklists is a killer feature, so definitely worth the effort to add that back in later.

PR #803, opened 2021-09-27
Switching forms to use editable select2 picklists after fixing issue with fitering and showing the insert select text first

brianbolt:
My previous attempt at this had 2 issues that were noted:

1) The picklists didn't insert the "insertSelectedOoption" as the first item in the picklist
2) The ability to filter was not working

I fixed both of these issues

### DIFF
--- a/modules/Components/src/client/PickList.coffee
+++ b/modules/Components/src/client/PickList.coffee
@@ -317,7 +317,10 @@ class PickListSelect2Controller extends PickListSelectController
 		if @options?.width?
 			@width = @options.width
 		else
-			@width = "100%"
+			# Select2 https://select2.org/appearance#container-width
+			# Uses the style attribute value if available, falling back to the computed element width as necessary.
+			@width = "resolve"
+
 		@propertyMap = acasPropertyMap
 		if options.propertyMap?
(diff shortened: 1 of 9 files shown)


modules/Components/src/client/PickList.coffee
```diff
@@ -343,9 +356,9 @@ class PickListSelect2Controller extends PickListSelectController
 		@rendered = true
 		@
 
-	addOne: (enm) =>
-# override to do nothing
-		return
+# 	addOne: (enm) =>
```
bffrost, line 359: nit: better to just remove than to leave dead code

bffrost (approved):
LGTM. Noted some optional changes, but OK with merging as-is.

brianbolt:
Even though we have many examples of setting the width through LsThing configurations, there was no place in the code where this was actually being passed down to the ACASForm controllers and therefore they were never actually passed to the PickLists or other elements.  I removed the `width` settings from ExampleThing to reflect that.  Additionally I set the select2 widget to `resolve` its width from its parent select element by default (https://select2.org/appearance#container-width).  I also added an example of how to override the width of the editable picklist items via CSS.

PR #808, opened 2021-10-20
ACAS SAML fails against Azure AD

### LINKED ISSUE #807: ACAS SAML fails against Azure AD
ACAS login fails if user already authenticated against AD with:
```
AADSTS75011: Authentication method 'Windowslntegrated' by which the user
authenticated with the service doesn't match requested authentication method
"Password, ProtectedTransport'. Contact the ACASDev application owner.
```
The is this:
https://docs.microsoft.com/en-us/troubleshoot/azure/active-directory/error-code-aadsts75011-auth-method-mismatch

 
The fix is to use the option: `disableRequestedAuthnContext: true`

```
disableRequestedAuthnContext: if truthy, do not request a specific authentication context. This is known to help when authenticating against Active Directory (AD FS) servers.
```

https://github.com/node-saml/passport-saml

brianbolt:
Fixes #807

### DIFF
--- a/app_template.coffee
+++ b/app_template.coffee
@@ -71,7 +71,7 @@ startApp = ->
 				entryPoint: config.all.server.security.saml.entryPoint
 				issuer: config.all.server.security.saml.issuer
 				cert: config.all.server.security.saml.cert
-				disableRequestedAuthnContext: true
+				disableRequestedAuthnContext: config.all.server.security.saml.disableRequestedAuthnContext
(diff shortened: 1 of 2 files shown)


brianbolt:
No, not specifically Okta. I'll check on that and report back.

brianbolt:
Too much effort to setup dev environment against okta so I just made it configurable with the default being `true`

PR #819, opened 2021-10-27
ACASFormLocationTree is broken in 1.13.6

### LINKED ISSUE #818: ACASFormLocationTree is broken in 1.13.6
We switched recently to an editable picklist (select2) controller.  This change broke the ability to pass the picklist without a URL.  ACASFormAbstractFieldController requires that you pass options to initialize now.

brianbolt:
Fixes #818


bffrost (approved):
LGTM

PR #834, opened 2021-12-02
Fix issue required to shows base64 images inline

brianbolt:
Fixes a few places we do this but discovered that Structure Images rendered in base64 in creg were broken without this.


bffrost (approved):
LGTM

PR #838, opened 2021-12-08
ACAS doesn't catch when a lot is not associated with a project

### LINKED ISSUE #837: ACAS doesn't catch when a lot is not associated with a project
ACAS allows customizable batch project services and should guard against the possibility that the service doesn't return a project for a lot.  

Additionally this bug exposed another bug with warning cross project loader role users when they are loading data against restricted compounds which are different than the experiment project.

brianbolt:
Fixes #837


bffrost (approved):
LGTM

PR #860, opened 2022-01-27
Brianbolt/issue859

brianbolt:


modules/Standardization/src/client/Standardization.html
```diff
@@ -66,6 +66,8 @@ <h3>Standardization Complete</h3>
                 <ul>
                     <li>Date of standardization: <span class="bv_standardizationCompleteRecordedDate"></span></li>
                     <li>Number of structures standardized: <span class="bv_standardizationCompleteStructuresStandardizedCount"></span></li>
+                    <li>Number of structured failed standardization: <span class="bv_standardizationCompleteStandardizationErrorCount"></span></li>
```
bffrost, line 69: typo: `structured` -> `structures`

bffrost (changes requested):
The whole radio button resolving to an array thing is weird, but I'm ok with it.

I didn't see where you made registration comment hidden by default. Is that in here?

Otherwise looks good aside from a typo.

brianbolt:
> I didn't see where you made registration comment hidden by default. Is that in here?

It was the standardization comment we wanted to hide but yeah, I hadn't hidden it by default. I just did that.

PR #862, opened 2022-02-02
Brianbolt/issue861

brianbolt:


jmcneil86043 (approved):
looks harmless, just adding a config

PR #870, opened 2022-02-17
Multiarch builds - acas

### LINKED ISSUE #869: Add multi-arch builds - acas

brianbolt:
Fixes #869

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -2,7 +2,7 @@ name: Docker Image CI
 
 on:
   push:
-    branches: [ release/1.13.7, release/1.13.6 ]
+    branches: [ release/* ]
 #   pull_request:
 #     branches: "**"
   create:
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,16 +13,14 @@ RUN \
 RUN   dnf install -y python36 python3-pip
 RUN   alternatives --set python /usr/bin/python3
 RUN   alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
-RUN		pip install argparse requests psycopg2-binary==2.8.6
-RUN   dnf install -y initscripts
+RUN		pip install argparse requests
+RUN   dnf install -y initscripts python3-psycopg2
```
brianbolt, line 17: psycopg2-binary wasn't working for me because they don't provide arm64 binaries yet, so instead installing by yum repo which works.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,55 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches: [ release/* ]
+#   pull_request:
```
bffrost, line 6: nit: if we're not planning on using it, let's remove this commented out code.

bffrost (approved):
LGTM

.github/workflows/docker-image.yml
brianbolt: Removed

PR #878, opened 2022-03-08
Brianbolt/issue491-simple

### LINKED ISSUE #491: Quotation marks and text after quotation marks are stripped in cmpdreg lot comments
Text after a quotation mark including the quote are stripped.  I.e. if you type: hello "test": your lot comments will be just be: hello I also verified with saving a lot with lots of !@#$%^&(*()_+μ`~<>?/.,|\}]{[

brianbolt:
Fixes #491
I also opened https://github.com/mcneilco/acas/pull/877 which also upgrades CmpdReg Backbone to the same version that ACAS uses.  However, this is probably too big of a fix to push right now so I created this new pull request.

- Update underscore from 2011 (1.1.6) to current ACAS underscore
- Use template "<%-" instead of "<%=" to escape html

I tested all of the relevant screens:

- Search
- Registration of new lot
- Update a lot
- Edit parent
- Add a new salt

For anyone with a LotView_Custom.inc page this fix requires a find and replace of `<%=` with `<%-`


bffrost (approved):
LGTM. @brianbolt can you make sure that the necessary changes to customer LotView_Custom.inc are documented somewhere as part of a 1.13.7 upgrade guide?

brianbolt:
> LGTM. @brianbolt can you make sure that the necessary changes to customer LotView_Custom.inc are documented somewhere as part of a 1.13.7 upgrade guide?

Yep.

PR #880, opened 2022-03-08
Brianbolt/issue879

### LINKED ISSUE #879: Purity operator should be null by default in single reg

brianbolt:
Fix requires equivalent updates to Lot_Custom.js on any acas_custom that has it.

Fixes #879

### DIFF
--- a/modules/CmpdReg/src/client/custom/Lot_Custom.js
+++ b/modules/CmpdReg/src/client/custom/Lot_Custom.js
@@ -168,6 +168,11 @@ $(function() {
           errors.push({'attribute': 'purity', 'message':  "Purity must be a number if provided"});
         }
       }
+      if (attr.purity!=null && attr.purity!='') {
+        if(attr.purityOperator==null || attr.purityOperator=='unassigned') {
+          errors.push({'attribute': 'purityOperator', 'message':  "Purity operator must be set if purity set"});
+        }
+      }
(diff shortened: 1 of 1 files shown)


modules/CmpdReg/src/client/custom/Lot_Custom.js
```diff
@@ -498,6 +507,9 @@ $(function() {
               if (this.purityMeasuredByCodeController.getSelectedModel().isNew()) purityMeasuredBy = null;
               else purityMeasuredBy = this.purityMeasuredByCodeController.getSelectedModel();
                 var vendor;
+              if (this.operatorCodeController.getSelectedModel().isNew()) purityOperator = null;
+              else purityOperator = this.operatorCodeController.getSelectedModel();
+                var tareWeightUnits;
```
bffrost, line 512: Weird indentation here for `var tareWeightUnits` and then it is not used directly below. Was there supposed to be a `var purityOperator` above here?

bffrost (changes requested):
Spotted one section that looks off

modules/CmpdReg/src/client/custom/Lot_Custom.js
brianbolt: Thanks, fixed this.

bffrost (approved):
LGTM!

PR #882, opened 2022-03-08
Brianbolt/issue881

### LINKED ISSUE #881: The 'Notebook' field should not be required when manually creating a new assay protocol
If I need to add a protocol to ACAS before I can upload an experiment that uses it, why is there not a 'Create new protocol' option in the protocol browser? If I'm using the 'Protocol Edit' page to make a new protocol, why are there required fields here that can be null in systems that don't require the protocol be entered to ACAS before loading experiment data? I can see why the Scientist field is helpful, but the 'Notebook' field is not relevant and it looks like users are just filling it with placeholder values that don't add any use.

brianbolt:
Fixes #881

Add configs for BaseEntity subclasses exeriment and protocol to override overall notebook requirement.
The default for experiment remains true and the protocol requirement has now been switched to false by default.

### DIFF
--- a/conf/config.properties.example
+++ b/conf/config.properties.example
@@ -431,6 +431,7 @@ client.experiment.lockwhenapproved.filter=${env.CLIENT_EXPERIMENT_LOCKWHENAPPROV
 #Note that notebook page can't be visible if notebook is not visible and page can't be required if notebook is not required
 client.entity.notebook.save=${env.CLIENT_ENTITY_NOTEBOOK_SAVE}
 client.entity.notebook.require=${env.CLIENT_ENTITY_NOTEBOOK_REQUIRE}
+client.entity.notebook.subclassExempt=${env.CLIENT_ENTITY_NOTEBOOKPAGE_REQUIRE}
 client.entity.notebookPage.save=${env.CLIENT_ENTITY_NOTEBOOKPAGE_SAVE}
 client.entity.notebookPage.require=${env.CLIENT_ENTITY_NOTEBOOKPAGE_REQUIRE}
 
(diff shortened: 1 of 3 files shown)


conf/config.properties.example
```diff
@@ -431,6 +431,7 @@ client.experiment.lockwhenapproved.filter=${env.CLIENT_EXPERIMENT_LOCKWHENAPPROV
 #Note that notebook page can't be visible if notebook is not visible and page can't be required if notebook is not required
 client.entity.notebook.save=${env.CLIENT_ENTITY_NOTEBOOK_SAVE}
 client.entity.notebook.require=${env.CLIENT_ENTITY_NOTEBOOK_REQUIRE}
+client.entity.notebook.subclassExempt=${env.CLIENT_ENTITY_NOTEBOOKPAGE_REQUIRE}
```
bffrost, line 434: I don't understand this line. What is `subclassExempt`?

modules/ServerAPI/src/client/BaseEntity.coffee
```diff
@@ -153,7 +153,9 @@ class BaseEntity extends Backbone.Model
 				saveNotebook= window.conf.entity.notebook.save
 			requireNotebook = true #default
 			if window.conf.entity?.notebook?.require?
-				requireNotebook= window.conf.entity.notebook.require
+				requireNotebook = window.conf.entity.notebook.require
+				if window.conf[attrs.subclass]?.notebook?.require?
```
bffrost, line 157: This makes sense. `client.entity.notebook.require` is the default setting for all entities, then if `client.protocol.notebook.require` or `client.experiment.notebook.require` are set, they override the default.

bffrost (changes requested):
Confused by `subClassExempt` but the rest looks good

PR #891, opened 2022-03-09
Barcode not saved properly in Creg single loader UI

### LINKED ISSUE #890: Barcode not saved properly in Creg single loader UI
## Expected Behavior
Lot details page saves a barcode entered

## Current Behavior
Barcode is not saved

## Steps to Reproduce

1. On the lot details page, add a barcode and save the lot
2. Navigate back to the lot and the barcode is not saved.
3. If supplier lot is not filled in the barcode will be filled into the supplier lot field

## Workaround
None

## Possible Solution
Update LotView_Custom.inc and replace:

```
				<input type="textfield" class="FormInput barcode" value="<%= supplierLot %>" /><br />
```

with:

```
				<input type="textfield" class="FormInput barcode" value="<%= barcode %>" /><br />
```

brianbolt:
## Description
Update LotView_Custom.inc had the model "supplierLot" instead of "barcode" mapped to it.

## Related Issue
Fixes #890

## How Has This Been Tested?
The barcode was being saved correctly on first save, but an update to the lot was setting the barcode to null because of this.
I tested by updating a lot and verifying that the barcode remained.

### DIFF
--- a/modules/CmpdReg/src/client/custom/LotView_Custom.inc
+++ b/modules/CmpdReg/src/client/custom/LotView_Custom.inc
@@ -89,7 +89,7 @@
             </div>
 			<div class="barcodeWrapper">
 				<label class="FormLabel">Barcode:</label>
-				<input type="textfield" class="FormInput barcode" value="<%= supplierLot %>" /><br />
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM

PR #895, opened 2022-03-11
/api/about route to self report version information Fixes #894

### LINKED ISSUE #894: Add a route to get version information
## Expected Behavior
ACAS should be able to report version information

## Current Behavior
No way to determine programmatically the ACAS version number

## Steps to Reproduce
NA

## Workaround
Look at docker image names on the server

## Possible Solution
Add version, build and buildTime to ACAS during builds and add the details to a configuration file so that ACAS can self report version information.

brianbolt:
## Description
/api/about is now a route in ACAS and can self report the version
 Example output:
```
{
    "acas": {
        "version": {
            "version": "release-1.13.7-versions",
            "buildTime": "2022-03-11T03:41:01.891Z",
            "revision": "8f0ace8ba1a0c0953c0f492aa6fd8e22e083990b"
        }
    }
}
```

I added an extra layer of `acas` so we can add racas and tomcat to this output if we want to in the future.

## Related Issue
Fixes #894 

## How Has This Been Tested?
Created branch `release/1.13.7-versions`
Let GitHub Actions build the new docker image
Started the new docker image and went to /api/about

/api/about reports
```
{
    "acas": {
        "version": {
            "version": "release-1.13.7-versions",
            "buildTime": "2022-03-11T03:41:01.891Z",
            "revision": "8f0ace8ba1a0c0953c0f492aa6fd8e22e083990b"
        }
    }
}
```


conf/acas-base.env
```diff
@@ -959,4 +959,13 @@ CLIENT_CMPDREG_BULKLOADSETTINGS_DBPROPERTIES=[{'name': 'CAS Number', 'dataType':
...
 # MAPS TO: client.cmpdreg.serverSettings.maxStandardizationDisplay
-CLIENT_CMPDREG_SERVERSETTINGS_MAXSTANDARDIZATIONDISPLAY=20000
\ No newline at end of file
+CLIENT_CMPDREG_SERVERSETTINGS_MAXSTANDARDIZATIONDISPLAY=20000
+
+# MAPS TO: client.version.version
+CLIENT_VERSION_VERSION=0.0.0
+
+# MAPS TO: client.version.buildTime
+CLIENT_VERSION_BUILDTIME=1970-01-01T00:00:00Z
```
brianbolt, line 968: This is iso formatted date

conf/acas-base.env
```diff
@@ -959,4 +959,13 @@ CLIENT_CMPDREG_BULKLOADSETTINGS_DBPROPERTIES=[{'name': 'CAS Number', 'dataType':
 CLIENT_CMPDREG_SERVERSETTINGS_LIVEDESIGN_PREPROCESSORSETTINGS=
 
 # MAPS TO: client.cmpdreg.serverSettings.maxStandardizationDisplay
-CLIENT_CMPDREG_SERVERSETTINGS_MAXSTANDARDIZATIONDISPLAY=20000
\ No newline at end of file
+CLIENT_CMPDREG_SERVERSETTINGS_MAXSTANDARDIZATIONDISPLAY=20000
+
+# MAPS TO: client.version.version
+CLIENT_VERSION_VERSION=0.0.0
```
brianbolt, line 965: Left this at semantic style versioning because I think we'll eventually go there with ACAS, this is just the default though incase its not set at build time.

package.json
```diff
@@ -68,6 +68,7 @@
     "request": "2.34.0",
     "request-promise": "4.2.1",
     "serve-favicon": "^2.5.0",
+    "semver": "7.3.5",
```
brianbolt, line 71: I added this package which is very popular because it lets you things like this:

```
if semver.lt('1.13.6, config.client.version.version) // true
  do something if version is less than 1.13.6
else
  do something if version is more than 1.13.6
```

We probably don't want to lean to hard into these types of checks in case we do switch to semantic versioning.  I added this for the case of CustomerSpecificServerFunctions and accounting for backwards compatibility.

conf/config.properties.example
```diff
@@ -731,3 +731,9 @@ client.cmpdreg.marvin.delayShowStep2Next=${env.CLIENT_CMPDREG_MARVIN_DELAYSHOWST
 client.cmpdreg.marvin.exportFormat=${env.CLIENT_CMPDREG_MARVIN_EXPORTFORMAT}
 client.cmpdreg.serverSettings.liveDesign.preprocessorSettings=
 client.cmpdreg.serverSettings.maxStandardizationDisplay=${env.CLIENT_CMPDREG_SERVERSETTINGS_MAXSTANDARDIZATIONDISPLAY}
+
+
+# About configs
+client.about.acas.version=${env.CLIENT_ABOUT_ACAS_VERSION}
```
bffrost, line 737: Something is odd to me about this being a config. If I change this config it has no effect on the version of the application - it would only cause the "about" endpoint to lie, right?

I can see that feeding this in as an environment variable during the build sets this config, and we wouldn't expect anyone to override it, but it still feels like a weird way of doing this.

bffrost (approved):

conf/config.properties.example
brianbolt: Thanks for the feedback, I guess I get that.  I don't see anyone overriding this and lying but I definitely see why its weird.  I think as long as we don't put too much functionality behind it right now, we can evolve the feature.

PR #897, opened 2022-03-15
Run acasclient tests on build

### LINKED ISSUE #896: Run acasclient tests on build
## Expected Behavior
ACAS should get tested with the tests from acasclient on build

## Current Behavior
No testing is done

## Steps to Reproduce
Na

## Workaround
Run acasclient tests manually.

## Possible Solution
Download acasclient, create the basic acas instance using docker compose and run the api tests.

brianbolt:
## Description

Main features
 - Startup basic acas oss instance, download the acasclient and run tests
    - If the build does not have equivalent branches in roo/rservices, it falls back to release/1.13.7 (using default git branch as a guide)
 - Build all branches but only push release and tags (previously built only tags)
 - Make roo wait for database to fully start
   - I was seeing some slow database startup times and it was causing an issue where roo wasn't coming up

See action build logs for this here: https://github.com/mcneilco/acas/actions/runs/1989388121

## Related Issue
Fixes #896

## How Has This Been Tested?
Built on a "release/" branch to make sure build test and push works
Built on a regular "brianbolt/" branch to make sure build and test works but does not push a docker image

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -2,48 +2,102 @@ name: Docker Image CI
 
 on:
   push:
-    branches: [ release/* ]
+    branches: ["**"]
   create:
     tags: "**"
 jobs:
   acas:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
-      -
-        name: Checkout
-        uses: actions/checkout@v2
-      -
-        name: Docker meta
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
         with:
           # list of Docker images to use as base name for tags
           images: |
             ${{ github.repository }}-oss
-      -
-        name: Set up QEMU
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-image.yml
```diff
@@ -2,48 +2,102 @@ name: Docker Image CI
...
 on:
   push:
-    branches: [ release/* ]
+    branches: ["**"]
   create:
     tags: "**"
 jobs:
   acas:
     runs-on: ubuntu-latest
-    strategy:
```
bffrost, line 11: What was the significance of removing this?

bffrost (approved):
LGTM. Not super familiar with the github actions syntax, but with your comments to explain the logic, this looks reasonable and looks right.

.github/workflows/docker-image.yml
brianbolt: > When jobs.<job_id>.strategy.fail-fast is set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

It was basically unnecessary to set it because we weren’t using matrix here.

Issue #902, opened 2022-03-18
Open Experiment in LiveDesign button should only add columns found in that experiment


bffrost:
## Expected Behavior
Open Experiment in LiveDesign button should only add columns found in that experiment

## Current Behavior
Open Experiment in LiveDesign button creates a LR with all columns associated with the protocol, regardless of whether they were part of this experiment.

## Steps to Reproduce
1. Load a new experiment onto any LD-ACAS server, associated with an existing protocol but with this SEL sheet having only one or two columns of results.
2. Click the Open in LiveDesign button in Experiment Browser
3. View the resulting LR

## Workaround
Users have been manually removing extra columns.

PR #903, opened 2022-03-18
Open Experiment in LiveDesign button should only add columns found in that experiment

bffrost:
## Description
Goal is to make Open in LiveDesign button only add columns found in that experiment, rather than adding all columns in that assay a.k.a. ACAS protocol.
When I went to do this, I found that this script was a bit of a mess, and I've learned something in my past 5 years of coding so I gave it a sizable refactor.
## Related Issue
#902 

## How Has This Been Tested?
Patched this file into a staging LD/ACAS server and confirmed clicking the button produces a LR with the correct compounds and _only_ the columns mentioned in the SEL sheet for that experiment.


brianbolt (approved):
I know you were waiting on some other feed back but LGTM so far.

bffrost:
@brianbolt I got my feedback, and the changes as-is are now final. I purposely branched this off `release/1.13.6` in case we wanted to backport it to 1.13.6. Any concerns with me merging both to `release/1.13.7` and `release/1.13.6`?

brianbolt:
My only concern about merging to 1.13.6 is a self inflicted one.  I am not sure all of the acasclient tests from the main branch of acasclient will pass against 1.13.6 so not sure if that build will work.  Happy to see it try though.  If not, I can always turn the dockerhub build for release/1.13.6 back on.

brianbolt:
Looks like no github actions was run on the initial branch or your merge to 1.13.6.  This is probably because the githubaction yaml file doesn't exist on those branches.  The tests did pass when running the 1.13.7 branch.  I'm going to go turn release/1.13.6 docker hub builds back on.

PR #904, opened 2022-03-18
Open Experiment in LiveDesign button should only add columns found in that experiment 

bffrost:
Copy of #903

Now merging into release/1.13.6


brianbolt (approved):
LGTM

Issue #905, opened 2022-03-18
ACAS-LiveDesign Integration: Open in LiveDesign button should retry if DI is still loading data


bffrost:
## Expected Behavior
If a user clicks the Open in LiveDesign button soon after loading an experiment, they still expect to be directed to a LR that contains all the compounds & data for that experiment.

## Current Behavior
The LiveReport creation will fail if DataIntegrator (DI) has not yet brought the new data from ACAS into LiveDesign and created assay columns that can be added to a LR

## Steps to Reproduce
1. Load an SEL sheet with a new assay name into an ACAS instance connected to LiveDesign
2. Immediately after uploading, click the Open in LiveDesign button (doing this _fast_ is key to reproduce)

## Workaround
Wait, click the button again

## Possible Solution
Retry several times within a fixed waiting period

PR #906, opened 2022-03-18
Open in LiveDesign button should retry if DI is still loading data

bffrost:
## Description
Added "wait_and_add_columns" utility which will repeatedly poll the column tree search until it finds desired results or times out.

## Related Issue
#905 

## How Has This Been Tested?
Patched this into a staging server, loaded an experiment into a new protocol w/ ~100 rows of data, clicked the Open In LiveDesign button very quickly. Observed that the page hung for ~20 seconds and then successfully redirected to a populated LR.


modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
```diff
@@ -153,29 +154,56 @@ def extract_endpoints_dict(node, endpoints_dict):
...
+    :param column_names: Full column names of format 'Foo (bar)'
+    :param max_retries: Maximum of retries, defaults to 10
+    :param delay_sec: Seconds of delay been tries, defaults to 5
+    """
+    success = False
+    for attempt in range(max_retries):
+        try:
+            col_ids = find_column_ids(
+                ld_client, assay_name, project_id=project_id, column_names=column_names)
+            if not col_ids:
```
brianbolt, line 180: I've seen issues where it takes longer for some data (I believe the dose response curves) to make their way to LD.  I think it would be better to do:
```
if len(col_ids) != len(col_names):
```

modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
```diff
@@ -153,29 +154,56 @@ def extract_endpoints_dict(node, endpoints_dict):
...
+                raise ValueError(
+                    f'Columns {column_names} under {assay_name} not found')
```
brianbolt, line 182: May need to adjust this message to say say something like:

```
'Some columns {column_names} under {assay_name} not found, found only these columns ids {col_ids}'
```

brianbolt (changes requested):
Looks really good except for 2 minor changes I think we need.

brianbolt (approved):
LGTM

PR #909, opened 2022-03-29
Remove need for extra docker environment files in acas #908

### LINKED ISSUE #908: Remove need for extra docker environment files in acas

brianbolt:
## Description
Requires updated mcneilco/acas-postgres:1.13.7 (https://github.com/mcneilco/acas-postgres/pull/4)

## Related Issue
Fixes #908 

## How Has This Been Tested?
I build the docker image and added it to my docker-compose.yml in base acas
I deleted on config/docker/**.env files
I ran on acasclient tests and they passed

### DIFF
--- a/conf/docker/acas/environment/acas.env
+++ b/conf/docker/acas/environment/acas.env
@@ -1,8 +1,0 @@
-ACAS=true
-APP_NAME=acas
-RUN_SYSTEM_TEST=false
-ACAS_DB_NAME=acas
-ACAS_SCHEMA=acas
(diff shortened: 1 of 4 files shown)


bffrost (approved):
LGTM.
@brianbolt do we want to consider explicitly changing the image tag for the db container?

brianbolt:
I don't think we want to do that right now.  I do see that being required in the future but this specific change is backwards compatible with all 1.13.7 and so no need right now.

PR #913, opened 2022-03-30
Open in Live Design button is broken on K8s

### LINKED ISSUE #912: Open in Live Design button is broken on K8s

brianbolt:
## Description
Add live design client url
If client url is provided, use it, otherwise fallback to the base url
I added "eprint" to print all non-essential output to stderr rather than stdout.  This makes it so we don't have to parse the stdout to get the url.  Parsing the stdout to get the url is tricky because there are 2 possible URLS that we could be looking for.  It seemed safer to me to avoid parsing the stdout at all and to assume that if the exit code was good, then the URL is the entirety of stdout.
Removed exec and replaced with the safer "spawn" exec command which is safer:

https://stackoverflow.com/questions/23697639/child-process-spawn-in-node-js-security-escaping
>If gm was using child_process.exec there would be a greater chance of injection. This is because child_process.exec executes the commands under a subshell and not directly, letting shell meta characters like backticks, $(), ;, &&, || etc to be used nefariously.

## Related Issue
Fixes #912

## How Has This Been Tested?
Ran on LD report creations with and without `client.service.result.viewer.liveDesign.clientUrl` config set.


bffrost (approved):
LGTM. Nice improvement here.

PR #925, opened 2022-04-08
Convert ModuleMenusConfiguration into a config

### LINKED ISSUE #914: Convert ModuleMenusConfiguration into a config
**Background**
One of the most common points of customization in ACAS is the menu on the left. Currently this is controlled by a coffeescript file called ModuleMenusConfiguration.coffee. This file essentially has configuration in it, and almost never contains any real javascript code / logic. In the effort to reduce the need for custom files in ACAS to ease ACAS k8s deployment.

**Scope of Work:**

We should convert the Module Menus configuration into a true ACAS config that can be specified via zzz-config.properties and ultimately from an environment variable.
Propose a design / JSON structure for the new config
Write a script to convert existing ModuleMenusConfiguration.coffee files into config
Make logic backwards-compatible so existing deployments with ModuleMenusConfiguration.coffee are not broken.

brianbolt:
## Description
 - Add new configuration `client.moduleMenus.menuConfigurationSettings`
 - Comment out `window.ModuleMenusConfiguration` (but leaving file behind as example)
 - In jade template, only pass `menuListJSON` if `window.ModuleMenusConfiguration` is defined (it's undefined now by default)
 - In ModuleMenus controller, use passed in `menuListJSON` if not null (it's null by default) and use `client.moduleMenus.menuConfigurationSettings` as a fallback
   
## Related Issue
#914 

## How Has This Been Tested?
 - Verified that in a default docker setup (`window.ModuleMenusConfiguration` is undefined), that the config is used and that changing it, results in an acas restart and the module menus changing. 
 - Verified that if I comment back in the ModuleMenusConfiguration.coffee file, thus defining `window.ModuleMenusConfiguration` that the coffee script configuration variable is used instead of the config


bffrost (approved):
LGTM. I am curious how the environment variable substitution works for things like `${CLIENT_ROLES_ACAS_USERROLE}` but I assume this was part of your testing so it must work somehow.

brianbolt:
> LGTM. I am curious how the environment variable substitution works for things like `${CLIENT_ROLES_ACAS_USERROLE}` but I assume this was part of your testing so it must work somehow.

Only so far as I know the final output of module menus includes what I have in `CLIENT_ROLES_ACAS_USERROLE` despite CLIENT_ROLES_ACAS_USERROLE being defined later in the file. SO that was good enough for me.

PR #931, opened 2022-04-19
Prepare Config files should override the order of configs using the order of config files readings

### LINKED ISSUE #930: Prepare Config files should override the order of configs using the order of config files readings
## Expected Behavior
ACAS should be able to run PrepareConfigFiles successfully when custom deployments add new configurations that is not in base ACAS which override configurations in base ACAS through interpolation.

```
deployName=blahblah
client.cmpdreg.clientUILabels.applicationNameForTitleBar=${deployName}
client.moduleMenus.homePageMessage=Welcome to ${deployName} ACAS
```

## Current Behavior
PrepareConfigFiles erros with:

```
acas_1       | Problem parsing /home/runner/build/conf/zzz-config.properties: Error: The property "deployName" does not exist
```

## Steps to Reproduce

1. Create a zzz config 
2. Add variables like in the description
3. ACAS runs prepare config files and errors like above

## Workaround
Hardcode the string in your variables

## Possible Solution

The reason this happens is because ACAS uses _.extend allConf, newConf when combining configs.  Because of this, existing configs like `client.cmpdreg.clientUILabels.applicationNameForTitleBar` are left at the top of the object and new configs like `deployName` are added to the end.  When ACAS parses the configs and tries to interpolate the ${deployName} variable, it is not yet defined.  The solution will be to mix in these variables but move existing configs to the bottom of the file.

brianbolt:
## Description
When overriding configs, delete key and add it back to the object rather than just replacing in place. This moves the key to the bottom and allows for interpolation

## Related Issue
Fixes #930

## How Has This Been Tested?
Added a zzz config like:

```
deployName=blahblah
client.cmpdreg.clientUILabels.applicationNameForTitleBar=${deployName}
```
Verified prepared config files worked instead of erorring
I printed the entire config before parsing and made sure `client.cmpdreg.clientUILabels.applicationNameForTitleBar=${deployName}` was moved to the bottom before ACAS tried to do interpolation. 
I verified that `client.cmpdreg.clientUILabels.applicationNameForTitleBar=blahblah` was set in the final configuration file.

### DIFF
--- a/modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
+++ b/modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
@@ -182,7 +182,14 @@ getProperties = (configDir) =>
 	allConf = []
 	for configFile in configFiles
 		console.info "reading conf file: #{configFile}"
-		allConf = _.extend allConf, propertiesParser.read(configFile)
+		conf = propertiesParser.read(configFile)
+		# Fill overwrite allConf with conf keeping conf variables at the bottom
+		# It's important that overwritten configs get moved to the bottom of the list
+		# as the variable interpolation is done in the order of the list
+		for key in Object.keys(conf)
+			if allConf[key] != undefined
+				delete allConf[key]
+			allConf[key] = conf[key]
 		console.info "read conf file: #{configFile}"
 
 	configString = ""


brianbolt:
Discovered that this change opens up additional issues like if you override a config which is re-used inside acas base, then by moving it to the bottom, you break the interpolation of that config.

PR #932, opened 2022-04-20
Maintain order when adding environment variable overrides

### LINKED ISSUE #928: ACAS_ environment variables
## Expected Behavior
ACAS should not require a secondary .env file to be configured with environment variables.
ACAS should allow environment variables prefixed with `ACAS_`  to replace configs

brianbolt:
## Description
Previously, environment variable overrides were appended to the bottom of a string of base configs, essentially duplicating the config but overriding the base config because it was last in.  However, if a separate base config used this config as part of its interpolation, it was still using the original, non-override.

Take for example base config `client.service.persistence.host=roo`

By using environment variable override:

```
ACAS_SERVICE_CMPDREG_PERSISTENCE_HOST=localhost
```

This successfully overrode  to `client.service.persistence.host=localhost`

But the dependent config:
```
client.service.persistence.fullpath=http://${client.service.persistence.host}:${client.service.persistence.port}/${client.service.persistence.path}/
```

Was interpolated BEFORE this override and thus was left to:

```
client.service.persistence.fullpath=http://roo:8080/acas/api/v1/
```

This change maintains the order of the configs as read from the config files.

## Related Issue
Fixes #928 

## How Has This Been Tested?
Verified compiled base configs from before and after this change had 0 diffs
Verified that setting `ACAS_SERVICE_CMPDREG_PERSISTENCE_HOST=localhost` overrode all dependent configs that used `client.service.persistence.host`


modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
```diff
@@ -185,16 +185,16 @@ getProperties = (configDir) =>
 		allConf = _.extend allConf, propertiesParser.read(configFile)
 		console.info "read conf file: #{configFile}"
 
-	configString = ""
+	confObject = {}
```
bffrost, line 188: I was concerned that a basic dict / object will not be guaranteed to retain order:
https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

It seems like a `Map` object might be a better choice for storing the configs and ensuring the order matches how they were read in.

bffrost (changes requested):

modules/BuildUtilities/src/server/PrepareConfigFiles.coffee
brianbolt: Thanks, I switched this over to use `Map`

bffrost (approved):
LGTM

Issue #933, opened 2022-04-20
Browsers should auto populate entry tables by default


bffrost:
## Expected Behavior
Admin modules like Assay Scientist and CmpdReg Scientist should auto-populate with results when the Browser module is first opened. This saves clicks and these searches are extremely fast.

## Current Behavior
Currently the browser is blank, and the user has to click into the search bar, type "*" then hit enter. Over time this is frustrating when using this module often.

## Steps to Reproduce
N/A

## Workaround
N/A

## Possible Solution
N/A

PR #934, opened 2022-04-20
CodeTable Admin module refactor and "searchOnLoad" feature

bffrost:
Significant refactor of CodeTable Admin modules to simplify and reduce repetition. Shifted CmpdRegAdmin modules onto CodeTable Admin forms as well by enhancing configurability. Also added new default behavior of searching for "*" immediately on load.

## Description
- Added AbstractCodeTable class which extends `Backbone.Model` and contains the string logic for expanding a `codeKind` into the displayName, pluralDisplayName, upperDisplayName, etc. This class also contains the validation logic that was repeated in many CodeTable-related Model classes.
- Enhanced the `AbstractCodeTablesAdminController` and `AbstractCodeTablesAdminBrowserController` controllers so they require fewer configs and draw more of their configs from their Model class. This is a breaking change that goes along with the new AbstractCodeTable class.
- Switched CodeTablesAssayScientist and CodeTablesCompoundScientist to use the new abstract classes, eliminating repeated code.
- Switched CmpdRegStereoCategory and CmpdRegVendor to use the same new abstract classes. This required adding a few more options for overriding urlRoot and deleteUrlRoot.
- Eliminated the now-unused AbstractCmpdRegAdmin modules. The AbstractCodeTable and AbstractCmpdRegAdmin classes were close copy/pastes of each other with slight differences in urls. This can now be accounted for with simple configuration on each implementation.
- Added a new feature to automatically search for "*" when an AbstractCodeTablesAdminBrowser instance loads, which saves users a click.

## Related Issue
#933

## How Has This Been Tested?
- Clicked through all 4 modules in question, ensured the regular search and the auto search worked.
- used the modules to create new entries, confirmed the new entries could be searched for in the browser
- deleted entries from each module and confirmed the deleted entries were no longer findable in the browser


modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminBrowser.coffee
```diff
@@ -174,13 +199,15 @@ class AbstractCodeTablesAdminBrowserController extends Backbone.View
...
+			urlRoot: @urlRoot
 			toDisplay: @toDisplay
 		@searchController.render()
 		@searchController.on "searchRequested", @handleSearchRequested.bind(@)
 		@searchController.on "searchReturned", @setupCodeTablesAdminSummaryTable.bind(@)
 		@searchController.on "createNewCodeTablesAdmin", @handleCreateNewCodeTablesAdminClicked.bind(@)
-	#@searchController.on "resetSearch", @destroyCodeTablesAdminSummaryTable
+		# If search on load, trigger a search for "*"
+		if @searchOnLoad
+			@searchController.doSearch('*')
```
brianbolt, line 210: 1. I think we should call `handleSearchRequested()` and then refactor handleSearchRequested a little bit to handle this new case.  The reason I say this is because `handleSearchRequested` is called programmatically in `handleBackToCodeTablesAdminBrowserClicked` and `handleConfirmDeleteCodeTablesAdminClicked` which could also benefit from this new behavior.

brianbolt: Unrelated to this PR but in scope, but we should also hide the `bv_codeTablesAdminControllerContainer` when doing `handleSearchRequested`  which we aren't doing right now:

i.e. add this to `handleSearchRequested`

```
@$(".bv_codeTablesAdminControllerContainer").addClass "hide"
```

brianbolt: Wondering if we could/should also add a "searchOnLoadDefaultSearchTerm" here.  i.e. I can imagine a code table search where the subclass wants to search for something other than "*" by default.  i.e. search by username or date? Just a thought and something to possibly hook into later even if it isn't perfect.

modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminBrowser.coffee
```diff
@@ -150,17 +150,42 @@ class AbstractCodeTablesAdminBrowserController extends Backbone.View
   		codeType
   		entityClass
   		entityControllerClass
+		htmlViewId
   		moduleLaunchName
   ###
```
brianbolt, line 155: Nit that the comment `###` is outdented here

modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminBrowser.coffee
bffrost: Thanks - this cleared up that bug.
I took all of your suggestions and added a new "defaultSearchTerm" that simply pre-populates the search box, visibly to the user.

I tested changing `defaultSearchTerm` and having it with & without `searchOnLoad` and all the combinations work.

brianbolt (approved):
LGTM

john-mft:
@bffrost On  my 2022.4.x of Nimbus I am getting this error on client load
CmpdRegVendor.js:7 Uncaught ReferenceError: AbstractCodeTable is not defined
    at CmpdRegVendor.js:7:24
    at CmpdRegVendor.js:25:4
    
I see that my index does include the includes for AbstractCodeTable.js and the other two. I can also see the AbstractCodeTable class in the JS console.

I think I may have a JS load-order problem. Nimbus' acas_custom does not override modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js

Did you or @brianbolt run into any of these issues when deploying this update?
Thanks

brianbolt:
Most of the instances I manage don't have a required client scripts. I think the required client scripts is overridden in your acas custom to include the required Marvin's Javascript so I think you need to keep it. To fix, this, I would just diff with the current version of acasclient scripts

https://github.com/mcneilco/acas/blob/49aa1ff2a269cc936c33c3d22facaaa92c147565/modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js#L52

john-mft:
Thanks Brian, Nimbus doesn't actually have a RequiredClientScripts either, that's the mystery I can't solve so far

brianbolt:
Ah sorry, I missed that you said it does not override it. I'm not sure what the issue is.

brianbolt:
I'd check if mcneilco private modules includes a required client scripts. It may need to be updated.

john-mft:
Thanks Brian, that was it.

PR #943, opened 2022-05-06
Allow SEL upload w/ parent alias values within lot corp name

### LINKED ISSUE #942: Allow SEL upload w/ parent alias values within lot corp name
Please implement an enhancement to ACAS that allows lot corp names to be formed from a parent alias and lot number, in additional support to the standard corp name and lot number.

brianbolt:
## Description
Adds a config `server.service.external.preferred.batchid.allowParentAliasLotNames`=true/false for allowing `Parent Alias Lot Names` to be used when requesting `Corporate Batch ID` preferred names in addition to actual  `Lot Corp Name`

`Lot Corp Name` format is defined by:
  - `client.cmpdreg.serverSettings.appendSaltCodeToLotName`= true/`false default` 
  - `client.cmpdreg.serverSettings.corpBatchFormat`= `corp_batch_saltcode` or `corp_saltcode_batch`

By default they are of the form:
 - CMPD-0000001-001
 
If `client.cmpdreg.serverSettings.appendSaltCodeToLotName` is set to true (which is rare), they could look like:
 - if `client.cmpdreg.serverSettings.corpBatchFormat`=`corp_batch_saltcode`
   - CMPD-0000001-001-HCl
 - or `client.cmpdreg.serverSettings.corpBatchFormat`=`corp_saltcode_batch`
   - CMPD-0000001-HCl-001

Additionally, the 3 configs listed here can play a role in how the lot corp names are configured.

```
client.cmpdreg.serverSettings.corpSeparator=-
client.cmpdreg.serverSettings.saltSeparator=
client.cmpdreg.serverSettings.batchSeparator=-
````

A `Parent Alias Lot Name` is formatted the same, except it replace `CMPD-0000001` with a `Parent Alias` e.g.:
 - `Asprin-001`

In order to match a `Parent Alias Lot Name`, the `Parent Alias` must be referenced by one and only one Parent.

## Related Issue
Fixes #942

## How Has This Been Tested?

- When 
  - `server.service.external.preferred.batchid.allowParentAliasLotNames=true`
  - AND
    -  One and only one Parent is found by Parent Alias
Screen Shot 2022-05-06 at 10 45 05 AM

 - When
   -  `server.service.external.preferred.batchid.allowParentAliasLotNames=true`
   - and
      - Parent Alias is not referenced by any Parents
      - or
      - Parent Alias is referenced by multiple Parents
Screen Shot 2022-05-06 at 11 47 32 AM

When
  - `server.service.external.preferred.batchid.allowParentAliasLotNames=false`
 
Screen Shot 2022-05-06 at 12 14 18 PM

Verified that lot corp names with no relation to the parent corp name or when salts are appended to the lot corp name, this still works:
Case where salt added:
Screen Shot 2022-05-10 at 7 39 56 AM
Manually updated the lot corp name using sql:
`update lot set corp_name = 'BRIANTEST-001' where corp_name =  'CMPD-0000001-002HCl';`
Result:
Screen Shot 2022-05-10 at 7 41 14 AM
 

Updated not registered message after PR feedback:
Screen Shot 2022-05-10 at 8 56 28 AM

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -559,9 +559,16 @@ validateCalculatedResults <- function(calculatedResults, dryRun, curveNames, tes
       }
     } else {
       for (row in 1:nrow(newBatchIds)) {
+        NOT_REGISTERED_MESSAGE <- paste0(mainCode, " '", newBatchIds$Requested.Name[row],
+                          "' has not been registered in the system.")
+        
+        # TODO: Put this behind the unique alias config when we add it to ACAS
+        if(applicationSettings$server.service.external.preferred.batchid.allowParentAliasLotNames && mainCode == "Corporate Batch ID") {
+          NOT_REGISTERED_MESSAGE <- paste0(NOT_REGISTERED_MESSAGE, " If the ", mainCode, " is a Parent Alias Lot Name it may also mean that the Parent Alias is referenced by zero or multiple Parents.")
+        }
+        NOT_REGISTERED_MESSAGE <- paste0(NOT_REGISTERED_MESSAGE, " Contact your system administrator for help." )
         if (is.null(newBatchIds$Reference.Code[row]) || is.na(newBatchIds$Reference.Code[row]) || newBatchIds$Reference.Code[row] == "") {
-          addError(paste0(mainCode, " '", newBatchIds$Requested.Name[row],
-                          "' has not been registered in the system. Contact your system administrator for help."))
+          addError(NOT_REGISTERED_MESSAGE)
         } else if (as.character(newBatchIds$Requested.Name[row]) != as.character(newBatchIds$Reference.Code[row])) {
           if (mainCode == "Corporate Batch ID" || inputFormat == "Gene ID Data") {
             warnUser(paste0("A ", mainCode, " that you entered, '", newBatchIds$Requested.Name[row],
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/GenericDataParser/src/server/generic_data_parser.R [hunk 1]
```diff
@@ -559,9 +559,16 @@ validateCalculatedResults <- function(calculatedResults, dryRun, curveNames, tes
       }
     } else {
       for (row in 1:nrow(newBatchIds)) {
+        NOT_REGISTERED_MESSAGE <- paste0(mainCode, " '", newBatchIds$Requested.Name[row],
+                          "' has not been registered in the system.")
+        
+        # TODO: Put this behind the unique alias config when we add it to ACAS
+        if(applicationSettings$server.service.external.preferred.batchid.allowParentAliasLotNames && mainCode == "Corporate Batch ID") {
+          NOT_REGISTERED_MESSAGE <- paste0(NOT_REGISTERED_MESSAGE, " If the ", mainCode, " is a Parent Alias Lot Name it may also mean that the Parent Alias is referenced by zero or multiple Parents.")
```
bffrost, line 567: I find this error message unnecessarily confusing. Most of the time I'd expect us to get an error purely because the parent alias isn't registered. The "zero or multiple" wording is especially confusing to me. I'm inclined to leave this expanded error message out, even though it would be helpful in that one scenario where a parent alias is duplicated across different parents.

bffrost (changes requested):

modules/GenericDataParser/src/server/generic_data_parser.R
brianbolt: Thanks, I'm just going to update this message per your feedback but leave it in.  The reason being that I have a TODO as part of the global parent alias uniqueness config to hide this message when that config is set to true as ACAS shouldn't allow multiple aliases.

PR #947, opened 2022-05-16
Enhance Open in LD UX Experience 

dalejerikson:
Implemented "Generate Link" button and "Copy Link" to clipboard button for experiments alongside the "Open In" button.

## Description

Added a new option that can generate a LR link and shows it to the user in a way that is easy to copy. Also added a "Copy Link" button to enable one-click copying. 

The goal of this is to click "generate link" and then wait for the Live Report link to be generated on the ACAS page instead of having the "Open In LiveDesign" selection take the user to a separate page that takes a long time to load. 

## Related Issue
Information for this can be found under SDGR [ACAS-272](https://jira.schrodinger.com/browse/ACAS-272)

This work also overlaps with [ACAS-277](https://jira.schrodinger.com/browse/ACAS-277). 

## How Has This Been Tested?
This has been tested using a stubbed LD route and LR link

### DIFF
--- a/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
+++ b/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
@@ -612,22 +612,8 @@ class ExperimentBrowserController extends Backbone.View
 
 	handleOpenInQueryToolClicked: =>
 		unless @$('.bv_openInQueryToolButton').hasClass 'dropdown-toggle'
-			if @experimentController.model.get('lsLabels') not instanceof LabelList
-				@experimentController.model.set 'lsLabels',  new LabelList @experimentController.model.get('lsLabels')
-			subclass = @experimentController.model.get('subclass')
-			if window.conf.entity?.saveInitialsCorpName? and window.conf.entity.saveInitialsCorpName is true
-				if @experimentController.model.get('lsLabels').getLabelByTypeAndKind("corpName", subclass + ' corpName').length > 0
-					code = @experimentController.model.get('lsLabels').getLabelByTypeAndKind('corpName', subclass + ' corpName')[0].get('labelText')
-				else if @experimentController.model.get('lsKind') is "study" and @experimentController.model.get('lsLabels').getLabelByTypeAndKind('id', 'study id').length > 0
-					code = @experimentController.model.get('lsLabels').getLabelByTypeAndKind('id', 'study id')[0].get('labelText')
-				else
-					code = @experimentController.model.get("codeName")
-			else if @experimentController.model.get('lsKind') is "study" and @experimentController.model.get('lsLabels').getLabelByTypeAndKind('id', 'study id').length > 0
-				code = @experimentController.model.get('lsLabels').getLabelByTypeAndKind('id', 'study id')[0].get('labelText')
-			else
-				code = @experimentController.model.get('codeName')
-
-			window.open("/openExptInQueryTool?experiment=#{code}",'_blank')
+			handleGetLinkQueryToolClicked()
+			window.open(@$('.bv_exptLink'),'_blank')
 
 	handleGetLinkQueryToolClicked: => 
 			# Preparatory Logic to Get Code to Pass to Generate Link Route 
--- a/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
+++ b/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
@@ -679,7 +679,7 @@ class ExperimentBrowserController extends Backbone.View
 	handleCopyLinkClicked: =>
     	link = @$('.bv_exptLink').value
     	if link? # Defined and not null
-        	alert("Link copied to clipboard!")
+			alert("Link copied to clipboard!")
     	else
 			alert("Unable to copy link to clipboard")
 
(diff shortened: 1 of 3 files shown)


dalejerikson:
Just wanted to get some feedback on this progress thus far before I continue

dalejerikson:
I can change the route I made to just return an error or general message if there is no LD instance connected instead of fetching the routes/links for other options

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -626,6 +629,42 @@ class ExperimentBrowserController extends Backbone.View
 
 			window.open("/openExptInQueryTool?experiment=#{code}",'_blank')
 
+	handleGetLinkQueryToolClicked: => 
+			# Preparatory Logic to Get Code to Pass to Generate Link Route 
+			if @experimentController.model.get('lsLabels') not instanceof LabelList
```
bffrost, line 634: I see this was copied from "handleOpenInQueryToolClicked". Can we refactor to DRY this up?

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -626,6 +629,42 @@ class ExperimentBrowserController extends Backbone.View
...
+			else
+				code = @experimentController.model.get('codeName')
+			# Add Generating Link Loading Mask 
+			@$('.bv_generatingLink').show()
+			# Call to Route to Get URL 
+			$.ajax
+				type: 'GET'
+				url: "/getLinkExptQueryTool?experiment=#{code}"
+				success: (response) => 
+					@$('.bv_getLinkResults').show()
```
bffrost, line 655: I think to satisfy both ACAS-272 and ACAS-277 you should merge the logic of `handleOpenInQueryToolClicked` and this `handleGetLinkQueryToolClicked`. Could we have both buttons trigger the same "get a link" logic, and then have the logic diverge to either open a new tab with that link or to display the copyable link?

bffrost (changes requested):

bffrost:
@dalejerikson the CI build failures are not your fault - they're due to my acasclient changes having been pushed, but your ACAS branch being behind my ACAS changes that go along with the acasclient changes. I am pretty sure if you merge `release/1.13.7` into your dev branch `Dale-ACAS-272` that it will fix the CI build failures.

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
dalejerikson: Of course!

brianbolt (changes requested):
I think you'll have an async issue with part of the code. I left a comment on that.

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -609,6 +612,11 @@ class ExperimentBrowserController extends Backbone.View
 
 	handleOpenInQueryToolClicked: =>
 		unless @$('.bv_openInQueryToolButton').hasClass 'dropdown-toggle'
+			handleGetLinkQueryToolClicked()
+			window.open(@$('.bv_exptLink'),'_blank')
```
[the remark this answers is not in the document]
dalejerikson: Thanks! Super clear! I'll test these approaches and see what leads to the best user experience :)

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -623,8 +627,55 @@ class ExperimentBrowserController extends Backbone.View
...
+					@$('.bv_exptLink').val(response)
+				error: (err) =>
+					console.log err
+				datatype: 'json'
+			# Take Away Generating Progress Mask 
+			@$('.bv_generatingLink').hide()
+
+	handleCopyLinkClicked: => 
+		@$('.bv_exptLink').select()
+		document.execCommand("copy")
```
brianbolt, line 677: I don't think we want to use this as it's using a deprecated browser feature:
https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

You should be able to replace this with:

```
handleCopyLinkClicked: =>
    link = @$('.bv_exptLink').value
    if link? # Defined and not null
            alert("Link copied to clipboard!")
    else
            alert("Unable to copy link to clipboard")
```

I don't really like us using `alerts` in this way but I think the alternative might be to complicated to implement right now.  Prefer to just show another element and then have it fade away after a few seconds.  Alerts are very in your face I think and generally not used unless we really need to block the user from something (i.e. from moving off the page because of unsaved data).

dalejerikson: Makes sense. Thanks for the background info. It really helps!

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -664,18 +666,22 @@ class ExperimentBrowserController extends Backbone.View
...
-	handleCopyLinkClicked: => 
-		@$('.bv_exptLink').select()
-		document.execCommand("copy")
-		alert("Link Copied to Clipboard!")
+	handleCopyLinkClicked: =>
+    	link = @$('.bv_exptLink').value
+    	if link? # Defined and not null
+            alert("Link copied to clipboard!")
+    	else
+			alert("Unable to copy link to clipboard")
```
brianbolt, line 684: Spacing here is off.

dalejerikson: Hmmm interesting my linter didn't catch this. Will re-push.

bffrost (approved):
LGTM

brianbolt:
LGTM

brianbolt (approved):

Issue #494, opened 2018-04-23
Experiment loader files with the same header but different data types will break the loader


brianbolt:
[DI_4a (1).xlsx](https://github.com/mcneilco/acas/files/1939540/DI_4a.1.xlsx)

Produces this roo error:

```Body was:
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE HTML SYSTEM "about:legacy-compat">
[...]

PR #949, opened 2022-05-17
ACAS-35: Fix SEL issue caused by file containing duplicate result type with different data types

bffrost:
## Description
#494 is a known issue that yields no warning, but causes the upload to hard fail within Roo. Fixed this by adding an explicit check in SEL validation that will prevent users from proceeding with files like this.

## Related Issue
#494 

## How Has This Been Tested?
Added acasclient unit test for this issue, confirmed they were failing before this fix, and confirmed tests pass after adding this fix.
See https://github.com/mcneilco/acasclient/pull/54


brianbolt (approved):
LGTM

PR #952, opened 2022-05-19
Dose response curve validation

brianbolt:
## Description
Validate dose response curves with configuration to completely disable the functionality.
New configs:
 - server.sel.validateDoseResponseCurves=true/false : turns on/off feature completely
 - Per rendering hint configuration of goodnessOfFit thresholds
   - Available stats: SSE, SST, rSquared
   - level of error (warning or error)

## Related Issue
ACAS-273

## How Has This Been Tested?
Verified on/off switch works to turn off curve validation
Verified validation is not run on Dose response module (meant for fitting curves)
Tested scenarios with a single csv file using 4 parameter D-R:
  - Unexpected flipped curve with positive slope violates stats
  - Max is set too high so should violate stats
  - Flat curve doesn't really fit the data, questionable stats
  - Good curve should generate no warning, or error
  - Missing all 4 parameters
  - Missing 2 parameters
  - Good curve filled in by fitted values
  - Missing slope
  - Missing Max
  - Missing Min
  - Missing EC50
  - Curve ID provided but no rendering hint
  - No curve id or rendering hint (no error)
  - Raw data points without matching curve id still produces correct error but doesn't break validation


modules/Components/src/client/BasicFileValidateAndSave.coffee
```diff
@@ -174,6 +175,7 @@ class BasicFileValidateAndSaveController extends Backbone.View
 		@notificationController.addNotifications(@errorOwnerName, json.errorMessages)
 		if json.results?.htmlSummary?
 			@$('.bv_htmlSummary').html(json.results.htmlSummary)
+			@$('.bv_htmlSummary .bv_doseResponseSummaryTable').dataTable()
```
bffrost, line 178: Will this line be unhappy if there is no `.bv_doseResponseSummaryTable`?

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -2908,9 +2908,330 @@ splitOnSemicolon <- function(x) {
...
+            stat <- names(goodnessOfFitParameters)[s]
+            # Get the stat variable value from the environment
+            statValue <- get(stat)
+            statValueText <- format(statValue, digits=3)
+            warnings <- list()
+            errors <- list()
+            for(r in 1:length(statThresholds)) {
+                threshold <- statThresholds[[r]]
+                class <- names(statThresholds)[r]
+                if(is.na(statValue) || !is.na(threshold$value) && !is.na(threshold$operator) && eval(parse(text = paste0('statValue ', threshold$operator, ' threshold$value')))) {
```
bffrost, line 3149: This whole section for each statThreshold is pretty dense. We're already 2 or 3 levels nested in a for loop, and we're doing both classification of curves and HTML formatting. 
I'd suggest shifting this portion out into a helper function that takes in the necessary info and yields an error + errorClass + title. That will keep this big nested for loop simpler and easier to read.

If you wanted to go a step further you could have this whole loop just return lists or tables of data (without HTML formatting) then do all the `<td>`, `<tr>` HTML formatting below.

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -2908,9 +2908,330 @@ splitOnSemicolon <- function(x) {
...
+        }
+        if(stat == mainCode) {
+          style <- paste0(' style="min-width: 140px;"')
+        }
+        if(stat == "curve id") {
+          style <- paste0(' style="min-width: 60px;"')
+        }
+        tr <- paste0(tr,sprintf('<th%s>%s</th>', style, stat))
+    }
+    thead <- paste0(thead,tr)
```
bffrost, line 3221: nit, optional: Could we use a library to do string interpolation to make this a little less paste-y?
https://stackoverflow.com/questions/30711019/better-string-interpolation-in-r

modules/Components/src/client/BasicFileValidateAndSave.coffee
brianbolt: No, this line was ok but admittedly I hadn't checked before this.  I just added a safety check anyway.

brianbolt: Made this just render the plain html table if it finds one and also added a warning so the user knows it couldn't be rendered as an interactive table.

modules/GenericDataParser/src/server/generic_data_parser.R
brianbolt: I did some of this.  I still have a lot of paste for the rows but I did implement some Markdown stuff which makes it a little cleaner looking.

modules/GenericDataParser/src/server/generic_data_parser.R
brianbolt: Still pasty but a little cleaner.

brianbolt:
I think I addressed all of the concerns here.

- Changes made
- Missing parameters render as warnings
- Only show R² in the table
- Show the Parameters in the table
- If missing parameters, show the same "parameters missing warning" in the table when moused over (rather than just saying R2 couldn't be calculated)
- If the data table can't be rendered for some reason then show plain HTML like this screenshot

Screen Shot 2022-06-02 at 5 41 12 PM

bffrost (approved):
LGTM

dalejerikson (approved):
👍

PR #953, opened 2022-05-19
ACAS-84: Update the "Project" field in "Protocol Editor".

iamanandkr:
#### Issue
Default option of the `Project` field on the Protocol Editor page is `Select Project` which when chosen means the protocol is available to all projects and not bound to one project and so the text is not truly reflecting the behaviour.

#### Fix
We change the text from `Select Project` to `Not Restricted` and add an info icon with a tooltip to explain the behaviour.

## How Has This Been Tested?
Protocol_Editor


bffrost (approved):
LGTM

PR #957, opened 2022-06-30
ACAS-328 shift endpoint behind /api so it does not require new proxy settings

bffrost:
## Description
When deployed in production, ACAS is behind an nginx proxy that only routes certain urls to ACAS. ACAS-277 added a new `/getLinkExptQueryTool` URL but we did not add it to the configuration service that controls that nginx proxy. A simpler solution is to move this route behind a pre-existing url beginning, namely `/api`.

After fixing initial issue, noticed and fixed 2 more issues with the copy link button.

## How Has This Been Tested?
Patched these into javascript on live server, tested features in UI


brianbolt (approved):
LGTM

PR #946, opened 2022-05-12
Secure and expose Author and Role management APIs

bffrost:
## Description
- Certain scripts or admins need the ability to manage ACAS authors and their roles. However this access is highly sensitive and only admins should be able to create new users or alter their roles via the REST API.
- In order to properly test these routes and the admin-vs-non-admin permissions, I have significantly refactored the 'SystemTest' nodeAPI routes to enable more flexible creation of "backdoor" user accounts in addition to "bob".
- Fixed a security bug where updateProjectRoles was not ensuring admin permissions.

## Related Issue
https://github.com/mcneilco/acasclient/pull/53

## How Has This Been Tested?
Test-driven development with acasclient tests. See https://github.com/mcneilco/acasclient/pull/53

PR #963, opened 2022-07-14
Fix logout, redirect issue and saml syncing issue

brianbolt:
### The bug
**Steps to Reproduce:** 
Click logout
**Expected Results:** 
User is logged out and redirected to the homepage
**Actual Results:** 
Clicking logout results in this error being displayed on a page to the user
```
Error: req#logout requires a callback function
--
at IncomingMessage.req.logout.req.logOut (/home/runner/build/node_modules/passport/lib/http/request.js:65:44)
at exports.logout (/home/runner/build/routes/loginRoutes.js:111:7)
at Layer.handle [as handle_request] (/home/runner/build/node_modules/express/lib/router/layer.js:95:5)
at next (/home/runner/build/node_modules/express/lib/router/route.js:144:13)
at Route.dispatch (/home/runner/build/node_modules/express/lib/router/route.js:114:3)
at Layer.handle [as handle_request] (/home/runner/build/node_modules/express/lib/router/layer.js:95:5)
at /home/runner/build/node_modules/express/lib/router/index.js:284:15
at param (/home/runner/build/node_modules/express/lib/router/index.js:365:14)
at param (/home/runner/build/node_modules/express/lib/router/index.js:376:14)
at Function.process_params (/home/runner/build/node_modules/express/lib/router/index.js:421:3)
```

**Extra Information:**
It look like our package.json had a mistaken setting of ">=0.4.1" which means we pick up any changes to the passport package.
In this case, we picked up version 0.6.0 which, among a bunch of security fixes, also made logout asynchronous.
Stack over flow of same issue: https://stackoverflow.com/questions/72336177/error-reqlogout-requires-a-callback-function
https://medium.com/passportjs/fixing-session-fixation-b2b68619c51d

Redirects after login also seem to be broken. i.e.
if I request http://localhost:3000/cmpdreg 
I am redirected to http://localhost:3000/login?redirect_url=http://localhost:3000/cmpdreg
After login I am not redirected to http://localhost:3000/cmpdreg

### The Fix

 - Lock package.json to current versions rather than `>` so we don't inadvertently pick up more changes
 - Fix logout by passing the next steps as a callback to req.logout
 - Fix redirects by adding `keepSessionInfo: true` to authentication
 - In the course of fixing this, we also found that saml auth was broken when sync users was true
   - Fix was to add back authorRoutes.syncUsers function[https://github.com/mcneilco/acas/pull/946](https://github.com/mcneilco/acas/pull/946)

## Related Issue
ACAS-341
ACAS-342

## How Has This Been Tested?
Ran manual tests for
 - Logout when using direct login method (not saml)
 - Login when using direct login method
 - Login when using saml authentication and sync roles true
 - When logged out, requesting an experiment path and a cmpd reg path and verifying I was properly redirected:
    - e.g. requested http://localhost:3000/experiment_base/codeName/EXPT-00000002
    - logged in
    - was properly redirected back to: http://localhost:3000/experiment_base/codeName/EXPT-00000002 after login


package.json
```diff
@@ -56,8 +56,8 @@
     "multer": "^1.4.2",
     "node-fetch": "3.2.6",
     "ncp": "2.0.0",
-    "passport": ">=0.4.1",
-    "passport-local": ">=1.0.0",
+    "passport": "0.6.0",
+    "passport-local": "1.0.0",
```
brianbolt, line 60: Just noting that 1.0.0 is the only version and what was currently installed.
And passport 0.6.0 is also the current version.

modules/Login/src/server/routes/loginRoutes.coffee
```diff
@@ -90,20 +90,20 @@ exports.changePost = (req, res) ->
...
-		redirectMatch = config.all.server.security.saml.logoutRedirectURL
-	else 
-		redirectMatch = req.originalUrl.match(/^\/logout\/(.*)\/?$/i)
-		if redirectMatch?
-			redirectMatch = redirectMatch[1]
-		else
-			if config.all.client.basePath?
-				redirectMatch = config.all.client.basePath
+	req.logout =>
+		if config.all.server.security.saml.use == true && config.all.server.security.saml.logoutRedirectURL?
```
brianbolt, line 94: Diff is a little funky but all I did was pass logout a function with `=>` and tab over the rest of the code here.

modules/ServerAPI/src/server/routes/AuthorRoutes.coffee
```diff
@@ -681,6 +681,28 @@ exports.updateAuthorAndRolesInternal = (author, user, callback) ->
 																		else
 																			callback null, response
 
+exports.syncRoles = (author, rolesToAdd, rolesToDelete, callback) =>
```
brianbolt, line 684: This is the function we removed in a preceding PR that broken saml sync roles. We just added it back here.

PR #966, opened 2022-07-19
ACAS-346 SSO and direct login redirect clean up (external link UI, host checking and better redirect consistenty)

brianbolt:
### Issue
**Steps to Reproduce**
On an ACAS server configured with SAML, when not logged in, try to navigate directly to the acas homepage, i.e. https://example.com/acas

**Expected Results**
Get redirected to Okta for login, then get redirected back to original ACAS url https://example.com/acas

**Actual Results**
Okta sign-in works, but user is sent to https://example.com/ which then redirects to LD homepage
If you then go to https://example.com/acas it works since you are logged in

**Extra Information**
Clicking the relevant ACAS app icon in Okta results in a smooth sign-in that takes you to the ACAS homepage.

### Description of fix
 - Direct login changes
   - Fixed the general issue above by not storing the redirect url in the session
   - Added redirect_login parameter to login post from login jade file
   - use general `exports.getRedirectUrl(req)` for determining redirect login url to set on login page
   - Added redirect notice to home page when redirecting a user to a page other than the home page
Screen Shot 2022-07-19 at 9 33 54 AM

   - Added host check to redirect_url
   - Added external link page if redirect_url somehow gets set to external host
Screen Shot 2022-07-19 at 9 48 11 AM
  
 - SSO login page changes
   - Refactored anonymous functions to named functions
   - Use same general `exports.getRedirectUrl(req)` function for determining relay state

## Related Issue
ACAS-346

## How Has This Been Tested?
- Ran redirect tests on localhost with ACAS login configuration where `client.basePath=/`
- Ran redirect tests on localhost with SAML login configuration where `client.basePath=/`
- Ran redirect tests on remote host with ACAS login configuration where `client.basePath=/acas`
- Ran redirect tests on remote host with SAML login configuration where `client.basePath=/acas`

### DIFF
--- a/modules/Login/src/server/routes/loginRoutes.coffee
+++ b/modules/Login/src/server/routes/loginRoutes.coffee
@@ -52,6 +52,8 @@ exports.loginPage = (req, res) ->
 	if req.user?
 		user = req.user
 
+	# Get an appropriate redirect URL for the user to be sent to after they log in.
+	# The login page adds this url to login post as a query parameter called redirect_url.
 	redirectUrl = exports.getRedirectUrl(req)
 
 	errorMsg = ""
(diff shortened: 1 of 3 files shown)


modules/Login/src/server/routes/loginRoutes.coffee
```diff
@@ -47,11 +34,28 @@ exports.setupRoutes = (app, passport) ->
...
 
+exports.getRedirectUrl = (req) ->
+	redirectUrl = null
+	
+	# Check to see if the RelayState value is set in the request
+	if req.body?.RelayState? && req.body.RelayState != ""
+		redirectUrl = req.body.RelayState
+		console.log "redirecting to #{redirectUrl}"
+	else
+		parsedUrl = url.parse(req.originalUrl || req.url)
```
bffrost, line 45: I'm having trouble understanding what this does, even after reading the docs https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost

Would help to provide some examples here of what the overall URL vs the pathname vs the path will produce

modules/ServerAPI/src/client/login.jade
```diff
@@ -25,6 +28,8 @@ block content
                                 input.btn.btn-primary(type="submit", value="Submit")
                                 - if (resetPasswordOption)
                                     a.forgotPasswordLink(href="/passwordReset") Forgot your password?
+        - if (redirectUrl)
+            div.span7.offset1.alert.alert-success.loginError After login your will be redirected to #{redirectUrl}
```
bffrost, line 32: typo: `your` -> `you`

modules/Login/src/server/routes/loginRoutes.coffee
brianbolt: It's a legacy API but it supports relative paths as opposed to the new WHATWG URL standard they added to node which forces URLs into a format which requires a host. I used it because its a standard nodejs library and gives me a quick way to grab the host, path and query parameters of a string. 

There is this nice diagram explaining what the pathname, host and path keys are on the object:

Screen Shot 2022-07-19 at 4 51 22 PM

PR #968, opened 2022-07-19
ACAS-350: Set version information via ACAS_ environment variables

brianbolt:
### Issue

**Steps to Reproduce:** 
 Deploy ACAS

**Expected Results:**
ACAS /api/about endpoint displays the version, buildTime and revision of ACAS

**Actual Results:***
 ACAS displays the default json

```
{"acas":{"version":"0.0.0","buildTime":"1970-01-01T00:00:00Z","revision":"UNKNOWN"}}
```

### Fix
 - Set `ACAS_` to beginning of environment variables to override defaults with build information. The GitHub actions build already was setting the variables but just needed to pass the correct names to have them override the configs.  

## Related Issue
ACAS-350

## How Has This Been Tested?
 - Looked at /api/about and verified settings were set correctly.


bffrost (approved):
LGTM!

PR #971, opened 2022-07-22
Fixed error downloading summary files for some .sdfs 

hhan-schrodinger:
**Please review ACAS-301 before this PR**

## Description
There was an issue in the CReg bulk loader where summary files for some .sdfs (with special characters) could not be downloaded before and after registration. Added URI encoding in file download links to fix. 

## How Has This Been Tested?
Tested using an .sdf file with special characters (model 310522#1 (1).sdf), was able to download summary files before and after registration. Also was able to download original .sdf, current .sdf, and summary files from the Purge Files page in the CReg bulk loader.

### DIFF
--- a/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
+++ b/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
@@ -954,7 +954,6 @@ class FileRowSummaryController extends Backbone.View
 			loadUser: @model.get('recordedBy')
 			currentFileLink: "/api/cmpdRegBulkLoader/getSDFFromBulkLoadFileID/" + reportID
 			currentFileName: currentFileName
-			fileLink: window.conf.datafiles.downloadurl.prefix + "cmpdreg_bulkload/" + fileName
 			reportName: reportName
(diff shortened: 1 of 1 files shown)


modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
```diff
@@ -956,7 +956,9 @@ class FileRowSummaryController extends Backbone.View
 			currentFileName: currentFileName
 			fileLink: window.conf.datafiles.downloadurl.prefix + "cmpdreg_bulkload/" + fileName
```
bffrost, line 957: Is `fileLink` in here twice? Looks like it

bffrost (changes requested):
Noticed a duplicate line. Other than that, looks good.

bffrost (approved):
LGTM!

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
hhan-schrodinger: @bffrost Good catch, thanks! Just fixed it.

PR #977, opened 2022-08-02
Add bulk download button to Experiment Browser

hhan-schrodinger:
## Description
Made download button that given a current search and filter term, creates and downloads a zip for all files for matching experiments. 
* Created /exportExperimentFiles in ExperimentBrowserRoutes.coffee that takes an array of file names, creates a .zip file from them, and returns the .zip file name
- Created downloadFiles function in ExperimentBrowser.coffee that extracts all current experiments that pass filtering in the table sends array of their associated files to /exportExperimentFiles, downloads a .zip of the file and notifies user 

## How Has This Been Tested?
Created several experiments with different attachments, and attributes. Also deleted experiments to make sure that deleted experiment files weren't being included in the downloaded files. 

Screen Shot 2022-08-02 at 4 47 21 PM

Screen Shot 2022-08-02 at 5 00 24 PM

### DIFF
--- a/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
+++ b/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
@@ -426,13 +426,6 @@ class ExperimentBrowserController extends Backbone.View
 				@serviceReturn = null
 			dataType: 'json'
 
-
-
-
-		
-
-
-
 	setupExperimentSummaryTable: (experiments) =>
 		@destroyExperimentSummaryTable()
 
--- a/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
+++ b/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
@@ -174,6 +174,8 @@ class ExperimentSimpleSearchController extends AbstractFormController
 
 		#create an array to store all experiment files
 		allExperimentFiles = []
+
+		#search for experiments
 		$.ajax
 			type: 'GET'
 			url: "/api/experiments/genericSearch/#{experimentSearchTerm}"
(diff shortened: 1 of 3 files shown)


modules/ExperimentBrowser/src/server/routes/ExperimentBrowserRoutes.coffee
```diff
@@ -5,8 +5,12 @@ Add this line to public/src/modules/ModuleMenus/src/client/ModuleMenusConfigurat
 
 ###
 
+path = require 'path'
```
bffrost, line 8: Does this get used? If not, should remove it. If it is getting used, it'd be better to move it into the route that's using it.

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -156,10 +156,208 @@ class ExperimentSimpleSearchController extends AbstractFormController
...
+		filterSearchTerm = $(".dataTables_filter input").val()
+
+		#replace "-" in filterSearchTerm with spaces (for filtering later on)
+		filterSearchTerm = filterSearchTerm.replaceAll("-", " ")
+
+		#create an array to store all experiment files
+		allExperimentFiles = []
+
+		#search for experiments
+		$.ajax
```
bffrost, line 179: I think there should be a way that we can avoid re-doing the search. Sometimes these searches can be quite slow, so it is worth investing a little effort here to seeing if we can avoid re-doing that.

I think we can get the list of Experiments from the ExperimentSummaryTable's `collection` attribute. Accessing that might require you moving this function onto the ExperimentBrowserController.
From "within" the ExperimentBrowserController, `@experimentSummaryTable.collection` should give the list of experiments.

I feel strongly that there should be a way to reuse the filtering capability built into the DataTables widget. You've done something clever here by reimplementing the filtering logic, but I'm concerned about subtle bugs where the filtering you've implemented doesn't quite match the filtering built into DataTables. It'd be much safer if we can access the DataTables object on the frontend and get the list of filtered rows out of it.

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
```diff
@@ -156,10 +156,208 @@ class ExperimentSimpleSearchController extends AbstractFormController
...
+									break
+																
+						#if the experiment passes the filtering criteria, append experiment files to total expeirment files
+						if passFiltering
+							for experimentFile in experimentFiles
+								allExperimentFiles.push experimentFile
+					catch error
+						console.log "There was an error parsing experiment metadata:" + error
+
+				#Get today's date to timestamp any experiment file exports
```
bffrost, line 310: From here on down looks good to me.

bffrost (changes requested):

modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee
hhan-schrodinger: I think that makes a lot of sense, I see how the current system is fragile and could be broken especially if the DataTable is modified or updated in the future. I wasn't aware of how the experiment data could be accessed that way, will try to re-do this using that instead!

modules/ExperimentBrowser/src/server/routes/ExperimentBrowserRoutes.coffee
hhan-schrodinger: Removed it!

bffrost (approved):
This looks great @hhan-schrodinger.

Please don't merge this, since we're planning to put it into the next release. But it looks good to go as it stands.

bffrost:
@hhan-schrodinger I've just created a temporary `release/2022.1.1` branch and switched your PR to target that, so it it's safe for you to merge this now.

PR #979, opened 2022-08-09
Fix issue with detection of failed login for applications that require 302 and location

brianbolt:
## Description
 - Some applications require 302 status and a location header set to /login when failing to login to ACAS, a change to the login routes returned 200 and did not set the location header

## Related Issue
ACAS-389

## How Has This Been Tested?
Tested against and application that requires 302 and location header set and verified this fixed the issue of detecting a failed authentication.
Verified that normal UI login worked as expected.


bffrost (approved):
LGTM

brianbolt:
This seems to have broken the feature of requesting a url while logged out and getting redirected to that URL post login.

PR #980, opened 2022-08-09
Logout redirect to base path

brianbolt:
## Description
 - /logout should redirect users to the configured base path, currently it's just redirecting to "/" which is LD
 - Going to /logout/cmpdreg should redirect users to /cmpdreg
    -  This is a feature added a long time ago to make links to /logout/cmpdreg easy to create
    - The only known use case is for the excel apps we created which call /logout/excelApps when logging users out but its still a use case.

## Related Issue
ACAS-390

## How Has This Been Tested?
Ran on Kubernetes instance of ACAS where login is /acas and verified logout takes you back to /acas
Ran on local instance of ACAS where login is / and verified logout takes you back to /
Ran /logout/cmpdreg on both K8s and local instance and verified login page is rendered with `After login you will be redirected to /cmpdreg` and you are then redirected back to K8s after login

### DIFF
--- a/modules/Login/src/server/routes/loginRoutes.coffee
+++ b/modules/Login/src/server/routes/loginRoutes.coffee
@@ -114,18 +114,20 @@ exports.changePost = (req, res) ->
 
 exports.logout = (req, res) ->
 	req.logout =>
+		# If using saml then redirect to the specified logout url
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM. Can't say I fully understand why this fix works, but the tests you describe doing seem like the right ones so I believe you this is the right fix.

PR #982, opened 2022-08-09
Set status code 401 when user is not authenticated and send the login page

brianbolt:
## Description
 - Applications using the ACAS api need a way to detect a failed login.  Instead of setting 302 and location /login. We should respond with a 401 and the login page.

## Related Issue
ACAS-389

## How Has This Been Tested?
Verified that normal UI login worked as expected.
Verified that requesting /cmpdreg when logged out rendered `After login you will be redirected to /cmpdreg` and that after successful login, the user was redirected to /cmpdreg.
Verified logout worked as expected.


bffrost (approved):
LGTM

PR #984, opened 2022-08-10
SSO login map group attributes to ACAS roles

brianbolt:
## Description
 - Switch to mapping specific group attributes to ls role configs which map the SSO group attribute to a system role
 - Add `ROLE_CMPDREG-READONLY` to users able to login (can search creg but nothing else)
 
Screen Shot 2022-08-10 at 7 52 26 PM

## Related Issue
ACAS-394

## How Has This Been Tested?
Verified features of ACAS were added or removed by adding roles to a user one by one.
 - Added user to acas-users group and verified they could login and see acas-user module menus but not admin or creg
 - Added user to acas-admins group and verified they could see the addition acas admin module menus but not creg 
 - Added user to cmpdreg-users and verified they could see creg but not creg admin
 - Added user to cmpdreg-admins and verified they could see cmpd reg admin modules and bulk load admin purge app
 - Revoked all roles except cmpdreg-readonly and verified I could login but not see any acas user, admin or creg admin modules but could search creg.
 - Verified that cross project loader role was given when assigned but didn't actually test the role functionality.

Verified as I added or removed user from groups, that the SQL below yielded the correct roles in the ACAS DB when they hit the login endpoint which synched their roles:

e.g.
```
SELECT a.user_name, r.ls_kind, r.role_name
FROM author a
join author_role ar on a.id=ar.author_id
join ls_role r on ar.lsrole_id=r.id
WHERE r.ls_type='System'
and a.user_name = 'brian.bolt@boltengineered.com';
```

Full access to all:
```
           user_name           | ls_kind |           role_name            
———————————————+————+————————————————
 brian.bolt@boltengineered.com | ACAS    | ROLE_ACAS-ADMINS
 brian.bolt@boltengineered.com | ACAS    | ROLE_ACAS-CROSS-PROJECT-LOADER
 brian.bolt@boltengineered.com | ACAS    | ROLE_ACAS-USERS
 brian.bolt@boltengineered.com | CmpdReg | ROLE_CMPDREG-ADMINS
 brian.bolt@boltengineered.com | CmpdReg | ROLE_CMPDREG-READONLY
 brian.bolt@boltengineered.com | CmpdReg | ROLE_CMPDREG-USERS
(6 rows)
```

### DIFF
--- a/modules/ServerAPI/src/server/CustomerSpecificServerFunctions.coffee
+++ b/modules/ServerAPI/src/server/CustomerSpecificServerFunctions.coffee
@@ -144,24 +144,21 @@ exports.isUserAdmin = (user) ->
 exports.findByUsername = (username, fn) ->
 	return exports.getUser username, fn
 
-formatSystemRolesFromSSOGroups = (ssoGroups) =>
-	# Formats the sso roles into ACAS roles
-	# Case insensitive check for "CMPDREG" in the name assigned the role to the lsKind 'CmpdReg'
-	# if not matched, it assigns to role to the lsKind 'ACAS'
-	roles = []
-	if ssoGroups?
-		if typeof ssoGroups == "string"
-			groups = [ssoGroups]
-		for group in ssoGroups
-			roleName = group.toUpperCase()
-			lsKind = 'ACAS'
-			if roleName.indexOf("CMPDREG") > -1
-				lsKind = 'CmpdReg'
-			roles.push
-				lsType: 'System'
-				lsKind: lsKind
-				roleName: "ROLE_#{roleName}"
+formatSystemRolesFromSSOProfile = (profile) =>
+	# Map Profile Attributes to ACAS/CmpdReg System Roles
+	config = require "#{ACAS_HOME}/conf/compiled/conf.js"
 	
+	# Parse the config which maps the profile attributes to system roles
+	profileGroupToACASRoles = JSON.parse(config.all.server.security.saml.profileAttributeToSystemRoles)
+	lsType = 'System'
+	roles = []
+	for profileGroup in profileGroupToACASRoles
+		if profile[profileGroup.ssoName]?
+			roles.push({
+				lsType: lsType
+				lsKind: profileGroup.lsKind
+				roleName: profileGroup.roleName
+			})
 	return roles
 
 exports.ssoLoginStrategy = (req, profile, callback) ->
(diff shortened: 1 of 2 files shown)


modules/ServerAPI/src/server/CustomerSpecificServerFunctions.coffee
```diff
@@ -144,24 +144,21 @@ exports.isUserAdmin = (user) ->
...
-				lsType: 'System'
-				lsKind: lsKind
-				roleName: "ROLE_#{roleName}"
+formatSystemRolesFromSSOProfile = (profile) =>
+	# Map Profile Attributes to ACAS/CmpdReg System Roles
+	config = require "#{ACAS_HOME}/conf/compiled/conf.js"
 	
+	# Parse the config which maps the profile attributes to system roles
+	profileGroupToACASRoles = JSON.parse(config.all.server.security.saml.profileAttributeToSystemRoles)
+	lsType = 'System'
```
bffrost, line 153: I missed the fact that this was "System" before. Originally I had thought the lsType was supposed to represent the origin of that role, so
"System": ACAS internal role management
"LDAP": from LDAP
"Project": for project roles (no longer used)
Could we change the lsType to "IdP" if that won't break anything? That will then keep it straight which roles came from where.

bffrost (approved):
LGTM, commented with an optional change. Please ignore if you think it'll break things.

modules/ServerAPI/src/server/CustomerSpecificServerFunctions.coffee
brianbolt: I actually didn't know this and there are several places in the code base where we've set "System" as the lsType.  I actually didn't realize the difference between LDAP and system until now and a little unsure how to proceed.  It's pretty easy to switch everything over to IdP from System, but when I do that, it does actually wipe out all other roles. i.e. I my user had System roles, then I switched everything to IdP and now my user only has IdP roles?

bffrost: I think that is expected. My understanding is that IdP or LDAP roles should take precedence over System roles. So some "internal" users or service accounts could have System roles, but if someone is getting any roles from IdP, we expect that user to get all their roles from the IdP.

bffrost: Resolving this conversation considering changing the lsType as out of scope.

PR #985, opened 2022-08-11
CReg Bulk Loader: Fixed sorting bug on "Has File Been Updated" column

hhan-schrodinger:
## Description
There was a bug where you could not sort "Has File Been Updated" on the purge files page in the CReg Bulk Loader. This was because the logic for checking for whether the file was updated or not was done and added to the HTML after the dataTable was already generated. 

To fix the bug, the checkForBulkLoadFileModifications() route for checking whether a file has already been updated or not was moved into the getFilesToPurge() route. As a result, the bulkFile data object contained data on whether the file had been modified earlier on, and there was no need to run a separate request to check and update the table after it had already been generated. 

## How Has This Been Tested?
Screen Shot 2022-08-11 at 4 51 19 PM
Screen Shot 2022-08-11 at 4 51 14 PM

Tested with local data.

### DIFF
--- a/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
+++ b/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
@@ -955,20 +955,12 @@ class FileRowSummaryController extends Backbone.View
 			currentFileLink: "/api/cmpdRegBulkLoader/getSDFFromBulkLoadFileID/" + reportID
 			currentFileName: currentFileName
 			reportName: reportName
+			bulkFileUpdated: @model.get('bulkFileUpdated')
 			#remove special characters from the links to prevent errors, but not from the displayed names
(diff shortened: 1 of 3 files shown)


bffrost (approved):
LGTM, thanks for merging these together.

We should keep an eye on the performance of this for servers with many large bulk load files, since this could get quite slow. In that case we may want to consolidate some of these many API calls, for example by creating a roo route that takes a big list of corpnames and returns the modified dates, or by modifying the `getLotsByBulkLoadFileID` route to return the modified date of each lot.

But let's see how this performs first as written.

hhan-schrodinger:
@bffrost I agree and imagine the performance of it as is is not optimal. When testing it with just the five or so bulk files (with only a a few compounds each), it takes a second or two to load the table. With a lot of files it will run slowly.

bffrost:
Closing this PR since we decided to remove this logic all together for the current release, due to performance issues observed by QA.

PR #986, opened 2022-08-12
Removed all "has file been updated?" parts of CReg Bulk Loader

hhan-schrodinger:
## Description
The "Has File Been Updated" column on the Purge Files window of the CReg Bulk Loader was loading to slowly. When moving this logic into another route, this caused the window to load slowly instead. The decision was made to hold off on adding this feature for the time being. 

## How Has This Been Tested?
Purge file window loads quickly without issue after removing "Has File Been Updated?" parts. 

Screen Shot 2022-08-12 at 2 06 47 PM


bffrost (approved):
LGTM, thanks for the quick turnaround.

PR #988, opened 2022-08-18
Add disableDeleteMyLots which dissallows anyone but cmpd reg users from deleting a lot

brianbolt:
## Description
Add new default config `client.cmpdreg.metaLot.disableDeleteMyLots` = `true` which disables deleting lots if the user is not a CregAdmin

## Related Issue
ACAS-408

## How Has This Been Tested?
Ran acas client tests
Verified with setting on true, creg user who owned the lot could not see delete button
Verified with setting on false, creg user who owned the lot could see the delete and use the delete button
Verified with either setting a creg admin could see the delete and use the delete button


bffrost (approved):
LGTM, this logic looks correct.

bffrost (approved):
LGTM!

PR #994, opened 2022-08-30
Open url utility function which workaround safari pop up blocker

brianbolt:
## Description
Added utility function to workaround Safari pop up blocker and used it in open experiment in query tool component.  This was technique being used in a few places but in slightly different ways:
https://github.com/mcneilco/acas/blob/ebc5cb75accb8f09ca249252685f76809b8d5696/modules/Standardization/src/client/Standardization.coffee#L75
https://github.com/mcneilco/acas/blob/65e0cdd4dc6a3fe920903c9a0652dcbfe79e1852/modules/ExperimentBrowser/src/client/ExperimentBrowser.coffee#L412
https://github.com/mcneilco/acas/blob/ebc5cb75accb8f09ca249252685f76809b8d5696/modules/Standardization/src/client/Standardization.coffee#L75
## Related Issue
ACAS-426

## How Has This Been Tested?
Click the open in live design button in both chrome and safari and made sure pop ups weren't blocked.
This does not make it so you can repeatedly click the open in live design button, it just makes the button work in Safari where it is currently getting blocked.

### DIFF
--- a/modules/Components/src/client/OpenExperimentInQueryTool.coffee
+++ b/modules/Components/src/client/OpenExperimentInQueryTool.coffee
@@ -33,7 +33,7 @@ class OpenExperimentInQueryToolController extends Backbone.View
                 success: (response) => 
                     # Take Away Generating Progress Mask 
                     @$('.bv_generatingLink').hide()
-                    window.open(response,'_blank')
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM

PR #996, opened 2022-08-31
ACAS-427: add uri encoding for datafiles urls

bffrost:
## Summary
We have a few places that we're serving up URLs based on a filename, and we're not properly URI encoding, so the links are broken if the filenames have special characters like %.

This fixes it.

## Testing performed
Patched this change in as javascript on a server demonstrating the issue, and confirmed it fixed the experiment file download link.

### DIFF
--- a/modules/Components/src/client/ACASFormFields.coffee
+++ b/modules/Components/src/client/ACASFormFields.coffee
@@ -867,9 +867,9 @@ class ACASFormLSFileValueFieldController extends ACASFormAbstractFieldController
 			if !displayText?
 				displayText = fileValue
 			if @displayInline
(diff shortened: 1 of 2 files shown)


brianbolt (commented):
When I copied a URL before this change it had replaced spaces with %20.  Is this already happening somewhere else, like on save, and this would fail if the file you tested with already had escaped spaces (double url encoded)?

dalejerikson (approved):
👍

brianbolt (approved):
LGTM.  I looked on a server with spaces I the file names and we aren't encoding them at all.  I also took a look through the code and didn't see a place where we encoded the file names so this looks good to me.

PR #997, opened 2022-09-01
ACAS-412: Partial work on rendering data column values within Experiment

bffrost:
## Description
I started to work on the rendering piece, and made some progress. I wanted to give you a chance to review this before shoving it all into your branch. I think this would be good to merge in tomorrow morning so we can build on top of it.

## How Has This Been Tested?
Tested locally that the "Add Endpoint" button in Protocol Editor still works
Tested viewing experiments both in Experiment Browser and Protocol Browser.

There's a known issue that the "units" dropdown starts out empty. I tested that I could use the "editable picklist" to add values to the database, and then they'd be correctly rendered when viewing saved experiments. But long term we'll need to change the backend logic to save these during the data loading, so it's OK to ignore this for now.

### DIFF
--- a/modules/ServerAPI/conf/ProtocolConfJSON.coffee
+++ b/modules/ServerAPI/conf/ProtocolConfJSON.coffee
@@ -247,6 +247,18 @@
 			,
 				typeName: "data column"
 				kindName: "column type"
+			,
+				typeName: "data column"
+				kindName: "column units"
+			,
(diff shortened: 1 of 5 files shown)


hhan-schrodinger (approved):
Looks good, I don't quite fully understand the adding a new endpoint logic yet, maybe we can talk about it later?

PR #1002, opened 2022-09-06
ACAS-6: Fix Maestro sketcher implementation

bffrost:
## Description
Switch Salt Browser sketcher to actually use Maestro sketcher when client.cmpdreg.sketcher = 'maestro', then fix a couple issues with instantiating the sketcher and getting & setting mols.

## How Has This Been Tested?
Tested locally by coping maestro sketcher code into dev environment, then ran with maestro sketcher configured. Confirmed I could register a salt, got the correct molecular weight and formula, could render the images, and the "edit salt" workflow started with the correct structure loaded in the sketcher.

In this local dev setup, I did observe an issue where the structure image wouldn't render during the "preview" phase before save. This appears to be a MOL V3000 issue with the indigo / cdk rendering code in Roo. 
```
acas-roo-1  | 06 Sep 2022 18:39:55 72605 [http-nio-8080-exec-3] ERROR com.labsynch.labseer.api.ApiStructureController [] - Caught exception in renderMolStructure
acas-roo-1  | org.openscience.cdk.exception.CDKException: This file must be read with the MDLV3000Reader.
acas-roo-1  |   at org.openscience.cdk.io.MDLV2000Reader.readAtomContainer(MDLV2000Reader.java:370) ~[cdk-io-1.5.13.jar:1.5.13]
acas-roo-1  |   at org.openscience.cdk.io.MDLV2000Reader.read(MDLV2000Reader.java:229) ~[cdk-io-1.5.13.jar:1.5.13]
acas-roo-1  |   at com.labsynch.labseer.service.StructureServiceImpl.readMolStructure(StructureServiceImpl.java:151) ~[classes/:?]
acas-roo-1  |   at com.labsynch.labseer.service.StructureServiceImpl.renderMolStructureBase64_aroundBody2(StructureServiceImpl.java:89) ~[classes/:?]
```
We may want to switch this to using CReg, or need to figure out how to send a V2000 MOL to the rendering service.

### DIFF
--- a/modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
+++ b/modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
@@ -31,6 +31,8 @@ exports.requiredScripts = [
 	// '/CmpdReg/marvinjs/gui/lib/promise-1.0.0.min.js',
 	// '/CmpdReg/marvinjs/js/marvinjslauncher.js',
 	// '/CmpdReg/client/custom/marvinStructureTemplate.js',
+	// The following line is for the Maestro sketcher
+	'/CmpdReg/client/schrodinger/maestrosketcher/maestrosketcherlauncher.js',
 	'/socket.io/socket.io.js',
 	'/lib/blockUI.js'
 ];
--- a/modules/ChemStructure/src/server/routes/ChemStructureRoutes.coffee
+++ b/modules/ChemStructure/src/server/routes/ChemStructureRoutes.coffee
@@ -206,7 +206,7 @@ exports.renderMolStructureBase64 = (req, resp) ->
 			else
 				console.log '--- line 153:   got ajax error trying to render molStructure'
 				console.log error
-				console.log json
+				console.log output
 				console.log response
 				resp.statusCode = 500
 				resp.end JSON.stringify "render molStructure failed"
--- a/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
+++ b/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
@@ -57,6 +57,7 @@ exports.setupRoutes = (app, loginRoutes) ->
 	app.get '/cmpdReg/parentLot/getAllAuthorizedLots', loginRoutes.ensureAuthenticated, exports.getAllAuthorizedLots
 	app.get '/cmpdReg/allowCmpdRegistration', loginRoutes.ensureAuthenticated, exports.allowCmpdRegistration
 	app.get '/cmpdReg/export/corpName/:lotCorpName', loginRoutes.ensureAuthenticated, exports.exportLotToSDF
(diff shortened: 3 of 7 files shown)


bffrost:
@dalejerikson I found a workaround / patch to fix the issue I mention above: https://github.com/mcneilco/acas-roo-server/pull/358

I am seeing slightly different behavior in the preview vs the image for the saved salt. For example, registering "HF" by just clicking the "F" atom and adding it yields an image just saying "F" for preview, but "HF" for the final one.

Preview:
Screen Shot 2022-09-06 at 12 22 03 PM

Final:
Screen Shot 2022-09-06 at 12 22 10 PM

dalejerikson:
Good looks good 

Not sure if that rendering issue is a blocker or not. I can test with more examples locally and see if it comes up again.

bffrost:
@dalejerikson I do think the rendering issue is serious, and it also uncovers the fact that we're using a different chemistry engine (CDK) to render images in this one module, which isn't good for consistency. This really is a CReg admin module, so it should use CReg services / the active CReg chemistry engine to render images.

I've made a new Roo PR to expose a route to do that: https://github.com/mcneilco/acas-roo-server/pull/359
And updated this PR to proxy that route through Node and use it instead.

bffrost:
I also noticed an issue locally where the Create Salt page would show me an error that the "Salt Name" was blank, even when I'd filled it in.
I tracked that down to the fact there were two div's called `bv_saltName`, both within scope of the SaltBrowserController, so it was selecting both and taking the first one (which was in an inactive page of the module). So I fixed that by renaming the class to a unique name.

dalejerikson:
> mcneilco/acas-roo-server#359

Ah that's interesting. Glad you were able to catch that. Thanks!

dalejerikson:
> I also noticed an issue locally where the Create Salt page would show me an error that the "Salt Name" was blank, even when I'd filled it in. I tracked that down to the fact there were two div's called `bv_saltName`, both within scope of the SaltBrowserController, so it was selecting both and taking the first one (which was in an inactive page of the module). So I fixed that by renaming the class to a unique name.

Ah that's interesting. Glad you were able to catch that. Thanks!

dalejerikson:
> @dalejerikson I do think the rendering issue is serious, and it also uncovers the fact that we're using a different chemistry engine (CDK) to render images in this one module, which isn't good for consistency. This really is a CReg admin module, so it should use CReg services / the active CReg chemistry engine to render images.
> 
> I've made a new Roo PR to expose a route to do that: [mcneilco/acas-roo-server#359](https://github.com/mcneilco/acas-roo-server/pull/359) And updated this PR to proxy that route through Node and use it instead.

That makes sense. I took the liberty of using another rendering route that was made available to me within ACAS Roo instead of using the main CReg Chemistry Engine to render structures. In my testing during implementation, I'm surprised I didn't see discrepancies like this. I'll try to follow patterns more consistently during future development. I reviewed that PR and it all looks good so I approved.

PR #1004, opened 2022-09-08
ACAS-304: Bugfix for saving issue

bffrost:
## Description
I took a sortof extreme approach to this, and tried to alter the EndpointController to use the ACASFormStateTableFormController, which was built to create a non-tabular layout, but still save values to a StateTable data structure.

It took a bunch of troubleshooting, but I was able to get that working with some minor tweaks to ACASFormFields and ACASFormStateTable and Protocol. With that, I was able to shift from our bespoke controllers into a more "LsThing-like" mode where we can configure `EndpointsValuesConf` and derive the fields from there.

The big advantage here is that we can then take advantage of pre-existing code that works properly for saving these values. I briefly had an issue where I was saving _two_ new values every time I made an edit, but I was able to track that down to there being multiple listeners, which counterintuitively I was able to fix by stopping calling "render". This is ultimately because the `ACASFormStateTableFormController` class has some overlap between its `initialize` and `render` functions, and they shouldn't both be called.

## How Has This Been Tested?
Local dev environment testing. Confirmed I'm able to
- Add new values to the dropdowns
- Save new rows
- Render back saved rows
- Edit empty rows and save values on the first try

### DIFF
--- a/modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
+++ b/modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
@@ -43,6 +43,8 @@ exports.applicationScripts = [
 	'/javascripts/src/Components/LSErrorNotification.js',
 	'/javascripts/src/Components/AbstractFormController.js',
 	'/javascripts/src/Components/AbstractParserFormController.js',
+	'/javascripts/src/Components/ACASFormFields.js',
+	'/javascripts/src/Components/ACASFormStateTable.js',
 	'/javascripts/src/Components/BasicFileValidateAndSave.js',
 	'/javascripts/src/Components/BasicFileValidateReviewAndSave.js',
 	'/javascripts/src/Components/ACASThingBrowser.js',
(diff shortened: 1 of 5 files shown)


bffrost:
@hhan-schrodinger I'm chasing one more issue which is that the columns have a confusing / buggy looking value in the dropdown at first. I'll update this PR when I figure out a fix.
Screen Shot 2022-09-08 at 4 24 56 PM

bffrost:
Ok, got that sorted:
Screen Shot 2022-09-08 at 4 33 43 PM

bffrost:
@hhan-schrodinger I also just flipped the Column Name and Data Type columns to "required", which shows a little "required" message when they are blank, which then goes away when they are filled. However you can still save the form (i.e. hit "Update") when these are blank. So there's still more work for you on the validation part.
Screen Shot 2022-09-08 at 4 35 20 PM

hhan-schrodinger (approved):
Looks good to me!

PR #1005, opened 2022-09-19
RM show() Edit Salt Okay Button

dalejerikson:
Okay button is part of another HTML element being shown/hidden so using .show()/.hide() on this specific button is not necessary

**Steps To Reproduce**

Edit a Salt in the Salt Browser. Go through the whole workflow so you end up back at the salt browser.

**Prior Bug Results**

There's a red "Okay" button below the browser results table which sticks around through any new actions you take in that module. 

**Fix Results**

Salt browser shows the same way as before your edit.


bffrost (approved):
LGTM

PR #1006, opened 2022-09-19
Updating/Fixing Error Message if No Salt Structure Provided

dalejerikson:
**Steps to reproduce:**

Try to register a salt without a structure.

**Prior to Bug Fix Results:**

The current workflow just abruptly ends after you click save with no notification of whether the salt registration is successful or not.
No error is displayed to the user about the structure being absent.

**Fixed Results:**

When you register a new salt without a structure, the UI should display an error that adding a structure is mandatory for registering the salt.

### DIFF
--- a/modules/ServerAPI/src/client/SaltBrowser.coffee
+++ b/modules/ServerAPI/src/client/SaltBrowser.coffee
@@ -270,7 +270,9 @@ class SaltBrowserController extends Backbone.View
 			console.log(saltName)
 			console.log(saltStruct)
 			# Need to Alert User of Empty Fields Somehow 
-			alert('Please fill in name, abbrev, and structure fields!')
+			@$('.bv_createNotifications').show()
+			@createNotificationController.clearAllNotificiations()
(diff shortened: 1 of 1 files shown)


dalejerikson:
Note: This PR Is to handle issues when a user either tries to create a salt without a structure OR if the user tries to edit an existing salt to have an empty/no structure.

modules/Components/src/client/ACASFormChemicalStructure.coffee
```diff
@@ -95,6 +95,12 @@ class MaestroChemicalStructureController extends Backbone.View
 	setMol: (molStr) ->
 		@maestro.sketcherImportText molStr
 
+	isEmptyMol: (molStr) ->
```
bffrost, line 98: hypernit / comment: Could also combine this with "getMol" so it doesn't need any input arguments. But I think this is A-OK the way it is.

bffrost (approved):
LGTM

PR #1007, opened 2022-09-20
ACAS-6 followup: expose CmpdReg Salts menu item to all CReg users

bffrost:
## Description
- Currently the "CmpdReg Salts" menu item is only exposed to ACAS users, and the "CmpdReg Admin" header above it is only exposed to CmpdReg Admins.
- I believe we actually want the CmpdReg Salts / Salt Browser exposed to all CReg users, including non-admins. So I've made that change and also changed the "CmpdReg Admin" header itself to be visible by all CReg users, just to provide a sensible header above the CmpdReg Salts menu item.

## How Has This Been Tested?
Tested locally.
As user with only "ROLE_CMPDREG-USERS", before this change I see:
Screen Shot 2022-09-20 at 7 43 05 AM
After the change I see
Screen Shot 2022-09-20 at 7 42 43 AM

As a user with only "ROLE_ACAS-USERS", before this change I see:
Screen Shot 2022-09-20 at 7 44 59 AM

After this change I see:
Screen Shot 2022-09-20 at 7 45 16 AM


brianbolt:
I feel like it's weird to have a menu items called `CmpdReg Admin` and expose it to non cmpdreg admins.  I'd be more inclined to create a new menu items called `CmpdReg` and put it under there.
Does the module hide the edit and delete buttons if the user is not a CmpdReg Admin?  If so I think we should add that or make it display a message that the user does not have the proper roles to edit and delete salts.

bffrost:
Closing this PR as there are some further changes I want to make

PR #1008, opened 2022-09-22
ACAS-449: Inf numeric values uploaded  to experiment loader should produce and error

brianbolt:
## Description
Present an error to the user if they upload Number values in experiment loader which are interpreted as Infinite.  Specifically, when the user uploads "Inf" or "-Inf".

## Related Issue
ACAS-449

## How Has This Been Tested?
Added acasclient test for experiment loader files with Inf and -Inf Number values


dalejerikson (approved):
🏴‍☠️ 👍

bffrost (approved):
LGTM

PR #1009, opened 2022-09-23
ACAS-462: Fix permissions for Salt Browser module

bffrost:
## Description
- CReg users should be able to view, download, and create salts. Shifted these calls to `/cmpdReg/` route calls, rather than `/api/cmpdRegAdmin/` so they don't require admin permissions
- Only CReg admins should be able to edit or delete salts. Fixed logic to only show these buttons when the user is a CmpdReg admin.

Also several Module Menus changes:
- Made CmpdReg Salts menu visible to cmpdreg users rather than acas users
- Renamed "CmpdReg Admin" module menu header to "CmpdReg Configuration" and made it visible by both users and admins
- Moved "Label Sequences" under "CmpdReg Configuration" since in practice that's only used for CReg, and also made it visible to cmpdreg admins rather than acas admins

## How Has This Been Tested?
Local testing with both cmpdreg user only and cmpdreg user + cmpdreg admin permissions
CmpdReg user sees:
Screen Shot 2022-09-22 at 11 45 04 AM

CmpdReg admin sees:
Screen Shot 2022-09-23 at 10 28 27 AM

Confirmed that users vs admins can do only the expected operations in the GUI


modules/ServerAPI/src/client/SaltBrowser.html
```diff
@@ -99,8 +99,8 @@ <h5>Searching for Salts containing text "<i class="bv_exptSearchTerm"></i>"</h5>
         </div>
         <div class="well well-small bv_saltControllerContainer hide" style="margin-top: 10px;">
             <div>
-                <button class="btn btn-danger bv_deleteSalt">Delete</button>
-                <button class="btn bv_editSalt pull-right">Edit</button>
+                <button class="btn btn-danger hide bv_deleteSalt">Delete</button>
```
bffrost, line 102: Hid these by default, so it simplifies the logic to only show them to admins. Also fixes a tiny UX issue where they'd flash visible then hidden again for CmpdReg Users

dalejerikson (approved):
All looks good and makes perfect sense :)

brianbolt (approved):
LGTM

PR #1010, opened 2022-09-28
ACAS-460 LsThingBrowser use flat advanced search results, fix hiding/showing of messages

brianbolt:
## Description
 - Use new "flat" advanced search returns and on click of a row, fetch the full ls thing object
 - Fix issue with dates values rendering as Nan-na-na
 - Fix issues with Search messages not rendering properly
 - Add notification of when max search results has been hit

## Related Issue
ACAS-460

## How Has This Been Tested?
ACAS client tests of new roo route changes
Manual testing of ACASThing browser with 10,000 things, clicking on rows, hitting max limit...etc.

### DIFF
--- a/modules/Components/src/client/ACASThingBrowser.coffee
+++ b/modules/Components/src/client/ACASThingBrowser.coffee
@@ -410,9 +410,9 @@ class ACASThingBrowserController extends Backbone.View
 		if response.maxResults <= response.numberOfResults
 			msg = "Browser results were limited to the first " + response.maxResults + " entries out of " + response.numberOfResults + " results"
 			@$('.bv_maxThingBrowserSearchResultsReached').html msg
-			@$("bv_maxThingBrowserSearchResultsReached").removeClass("hide")
+			@$(".bv_maxThingBrowserSearchResultsReached").removeClass("hide")
 		else 
-			@$("bv_maxSearchResultsLimited").addClass("hide")
+			@$(".bv_maxThingBrowserSearchResultsReached").addClass("hide")
 
 		if results is null
 			@showOne("bv_errorOccurredPerformingSearch")
--- a/modules/Components/src/client/ACASThingBrowser.coffee
+++ b/modules/Components/src/client/ACASThingBrowser.coffee
@@ -69,16 +69,39 @@ class ThingSimpleSearchController extends AbstractFormController
 					values: [
 
 					]
-
+				returnDTO: [
+					
+				]
+									
 			queryTerms = @getQueryTerms(defaultQueryTerms, thingSearchTerm)
+
+			for queryValue in @configs
+				valDef = @model.getValueInfo(queryValue.key)
+				if valDef?
+					queryTerms.returnDTO.push	
+						stateType: valDef.stateType
+						stateKind: valDef.stateKind
+						valueType: valDef.type
+						valueKind: valDef.kind
+						key: queryValue.key
+				else
+					# If the key is an ls label
+					labDef = @model.getLabelInfo(queryValue.key)
+					if labDef?
+						queryTerms.returnDTO.push	
+							labelType: labDef.type
+							labelKind: labDef.kind
+							key: queryValue.key
+
+
 			$.ajax
 				type: 'POST'
-				url: "#{@genericSearchUrl}#{@model.get("lsType")}/#{@model.get("lsKind")}?format=nestedfull"
+				url: "#{@genericSearchUrl}#{@model.get("lsType")}/#{@model.get("lsKind")}?format=flat"
 				dataType: 'json',
 				contentType: 'application/json'
 				data: JSON.stringify(queryTerms)
-				success: (thing) =>
-					@trigger "searchReturned", thing.results
+				success: (response) =>
+					@trigger "searchReturned", response.results
 				error: (result) =>
 					@trigger "searchReturned", null
 				complete: =>
@@ -336,23 +359,22 @@ class ACASThingBrowserController extends Backbone.View
 		@searchController.on "searchReturned", @setupThingSummaryTable.bind(@)
 		@$('.bv_queryToolDisplayName').html window.conf.service.result.viewer.displayName
 
-	setupThingSummaryTable: (things) =>
+	setupThingSummaryTable: (flattenedThings) =>
 		@destroyThingSummaryTable()
 
 		@$(".bv_searchingThingsMessage").addClass "hide"
-		if things is null
+		if flattenedThings is null
 			@$(".bv_errorOccurredPerformingSearch").removeClass "hide"
 
-		else if things.length is 0
+		else if flattenedThings.length is 0
 			@$(".bv_noMatchingThingsFoundMessage").removeClass "hide"
 			@$(".bv_thingTableController").html ""
 		else
 			@$(".bv_searchThingsStatusIndicator").addClass "hide"
 			@$(".bv_thingTableController").removeClass "hide"
-			thingCollection =  Backbone.Collection.extend 
-				model: @modelClass
+
 			@thingSummaryTable = new ThingSummaryTableController
-				collection: new thingCollection things
+				collection: new Backbone.Collection flattenedThings
 				configs: @configs
 				columnFilters: @columnFilters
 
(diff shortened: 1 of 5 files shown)


modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -69,16 +63,48 @@ class ThingSimpleSearchController extends AbstractFormController
 					values: [
 
 					]
-
+				returnDTO: {
+					thingValues: []
+					thingAttributes: ["id", "codeName"]
```
brianbolt, line 68: We need codeName and id for the ls thing editor forms when you click on a row. So we always pass these in the Ajax request.

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -69,16 +63,48 @@ class ThingSimpleSearchController extends AbstractFormController
...
+						queryTerms.returnDTO.thingValues.push	
+							labelType: labDef.type
+							labelKind: labDef.kind
+							key: queryValue.key
+
+
 			$.ajax
 				type: 'POST'
-				url: "#{@genericSearchUrl}#{@model.get("lsType")}/#{@model.get("lsKind")}?format=nestedfull"
+				url: "#{@genericSearchUrl}#{@model.get("lsType")}/#{@model.get("lsKind")}?format=flat"
```
brianbolt, line 102: new flat format request

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -69,16 +63,48 @@ class ThingSimpleSearchController extends AbstractFormController
...
 				dataType: 'json',
 				contentType: 'application/json'
 				data: JSON.stringify(queryTerms)
-				success: (thing) =>
-					@trigger "searchReturned", thing.results
+				success: (response) =>
+					@trigger "searchReturned", response
```
brianbolt, line 107: Instead of only passing the "results" we pass the full object so we can use the maxResults and numberOfResults to display a message to the user about hitting the limit of results.

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -163,15 +189,34 @@ class ACASThingBrowserCellController extends Backbone.View
...
+					content = @model.escape(@configs.key)
 		else
-			content =  @model.escape(@configs.key)
+			# This is strictly a key/pair object
+			content = @model[@configs.key]
+
+		if @options.thingModelExample?
+			modelValue = this.options.thingModelExample.get(@configs.key)
+			if modelValue instanceof Value && modelValue.get("lsType") == "dateValue" && !@configs.formatter?
+				content = UtilityFunctions::convertMSToYMDDate(content)
```
brianbolt, line 219: There is a lot of magic here, but this is all about determining the type of response (Thing, plain backbone model, or plain javascript key/pair object) and getting their keys.  One of the huge slowdowns of this entire thing was that we were creating a collection of thousands of LsThings which are really slow to instantiate. The default Browser now works using plain javascript objects in the rows (flat response),  but it's possible to pass lsthings of Backbone objects to this class as well (without search browser controller).

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -163,15 +189,34 @@ class ACASThingBrowserCellController extends Backbone.View
...
+				if typeof(content) == "string"
+					content = @model.escape(@configs.key)
 		else
-			content =  @model.escape(@configs.key)
+			# This is strictly a key/pair object
+			content = @model[@configs.key]
+
+		if @options.thingModelExample?
+			modelValue = this.options.thingModelExample.get(@configs.key)
+			if modelValue instanceof Value && modelValue.get("lsType") == "dateValue" && !@configs.formatter?
```
brianbolt, line 218: This particular line is important for date values.  We now pass one default plain instantiated ls thing to this method so that we can look up the keys and determine if we should format the returned values as dates.

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -331,42 +380,83 @@ class ACASThingBrowserController extends Backbone.View
...
+			c = new @modelClass(thing.attributes)
+			c.fetch
+				success: (model, response, options) =>
+					@thingController = new @controllerClass
+						model: model
+						configs: @configs
+						el: @$('.bv_thingController')
+					@renderController()
+				error: (model, response, options) =>
+					@$('.bv_thingController').html "Error fetching thing"
```
brianbolt, line 457: When a row is clicked, if the row is not a Thing instance (it's a plain javascript object by default now), then we use the row (which includes the codeName and id) and fetch it from the server and pass it to the form editor controller.

modules/Components/src/client/ACASThingBrowserView.html
```diff
@@ -71,14 +71,15 @@ <h5>Searching for <%-thingName%>s containing text "<i class="bv_exptSearchTerm">
...
+                        <h5 class="bv_maxThingBrowserSearchResultsReached hide"></h5>
                     </td>
                 </tr>
                 </tbody>
                 </tbody>
             </table>
         </div>
-
         <div class="bv_thingTableController hide" style="margin:0px; padding:10px; overflow: auto; overflow: auto;"></div>
+        <label class="bv_maxThingBrowserSearchResultsReached" style="margin-left:10px;"></label>
```
bffrost, line 82: Is this same `bv_maxThingBrowserSearchResultsReached` intended to be in here twice?

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -331,42 +380,83 @@ class ACASThingBrowserController extends Backbone.View
...
 			@thingSummaryTable.on "selectedRowUpdated", @selectedThingUpdated
 			@$(".bv_thingTableController").html @thingSummaryTable.render().el
 
 	selectedThingUpdated: (thing) =>
 		@$('.bv_thingControllerWrapper').append("<div class='bv_thingController'></div>")
 		@trigger "selectedThingUpdated"
-		@thingController = new @controllerClass
-			el: @$('.bv_thingController')
-			model: thing
-			readOnly: true
```
bffrost, line 368: I don't see `readOnly` in your new split implementation below. I think that is important to pass, right?

bffrost (approved):
This is pretty complex. I want to give @dalejerikson a chance to review and digest this.

I noticed a few places where we might've left out something important, but otherwise this looks like a nice cleanup.

modules/Components/src/client/ACASThingBrowser.coffee
brianbolt: Thanks.  I did not remove this on purpose. It's strange it's hardcoded to true but I'm not fixing that here.

brianbolt: Fixed by adding it back

modules/Components/src/client/ACASThingBrowserView.html
brianbolt: Thanks, removed the extra one

modules/Components/src/client/ACASThingBrowser.coffee
```diff
@@ -57,6 +50,7 @@ class ThingSimpleSearchController extends AbstractFormController
 			defaultQueryTerms =
 				queryString: "#{thingSearchTerm}"
 				queryDTO:
+					maxResults: @options.maxResults
```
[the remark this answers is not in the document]
dalejerikson: Is there a "maximum" maxResults limit? Wondering what stress tests look like.

dalejerikson (approved):
Makes sense to me! Helped to look at https://github.com/mcneilco/acas-roo-server/pull/367 first. Excited about the performance increase! 🥇

bffrost:
@dalejerikson Since these changes are really to enable your work, let's hold off on merging this PR into the release until you've gotten a chance to test out this module within your code. That way if you find other small things to tweak, we can fix them on this branch and avoid temporarily breaking anything in the release branch.

dalejerikson:
> @dalejerikson Since these changes are really to enable your work, let's hold off on merging this PR into the release until you've gotten a chance to test out this module within your code. That way if you find other small things to tweak, we can fix them on this branch and avoid temporarily breaking anything in the release branch.

Alright sounds good. I'll try to find some time today in between meetings and test this out to observe behavior.

modules/Components/src/client/ACASThingBrowser.coffee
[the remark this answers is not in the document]
brianbolt: I missed your comment here @dalejerikson . I tasted with 10,0000 and it was fairly performant depending on the needs. I think you'll have to play around with it and find the acceptable limit.

dalejerikson:
When implementing this in the context of CompBio I noticed the functionality working well. The performance increase is also really nice. However, this might break some parts of `ACASThingBrowserController`. I noticed the following when running this branch locally... 

Can you confirm if this is true on your end or if this a misconfiguration mistake on my part? 

Screen Shot 2022-10-17 at 12 18 14 PM

brianbolt:
Thanks @dalejerikson just pushed a fix for this and 2 other issues:

1) Fix for the thing browser controller: This was because thing.attributes didn't exist.  I had made a change at the last minute to pass the plain javascript key/pair object to this method rather than a generic backbone model (non thing backbone model) which would have attributes and I broke the creation of the model.
2) I fixed the Edit button to open the deep link to example things (see change to ControllerRedirectConf.coffee)
3) Fixed the readonly portion of the browser to not lock the example thing when in readonly mode in the browser.

dalejerikson:
Thanks @brianbolt , I'll update the CompBio Server, do some testing and see if there's any additional abnormal experiences.

brianbolt:
Nothing further from me. Thanks for testing it out and finding some issues.  Merging now.

PR #1014, opened 2022-09-30
ACAS-271: Preserve file naming in SDF bulk loader

bffrost:
## Description
ACAS introduces spaces and numbers in parentheses to keep SDF and zip filenames distinct because it drops them all into a common privateUploads folder. This makes life harder for users who may upload the same file multiple times but get different filenames when they re-download that same file from ACAS, or when they download derived zipfiles.

This acas change adds an "originalFileName" variable which is tracked and passed to the roo layer. Across the validation, registration, and purge pages, these changes make it so we show the original filename (`example.sdf`) and download as the original filename (`example.sdf`), but href to the correct filename (`example (26).sdf`) in cases where the same filename has been uploaded many times.

These changes are co-dependent with https://github.com/mcneilco/acas-roo-server/pull/370

## How Has This Been Tested?
Local testing of uploading the same file many times. Confirmed that the validation download zip, its contents, the registration download zip, and all the links on the purge files page (including the link / summary after purging) all show the original filename, and all download the correct numbered / distinct file.


brianbolt (approved):
LGTM

PR #1015, opened 2022-10-04
Add more descriptive error message when user attempts to delete Stereo Category and Vendors with dependencies

dalejerikson:
**Steps to Reproduce:**

_Stereo Category_

CmpdReg Stereo Categories >> Select a Stereo Category which has dependencies (e.g. compounds, compound lots etc. registered under it) 
Attempt to 'Delete' the category 

_Vendors_

CmpdReg Vendors >> Select a Vendor which has dependencies (e.g. compounds, compound lots etc. associated with it) 
Attempt to 'Delete' the vendor

**New Implementation Results:**

_Stereo Category:_ The Stereo Category is not deletable & the action throws an error, describing the exact cause of the error 

_Vendor:_ The Vendor is not deletable & the action throws an error, describing the exact cause of the error

**Previous Results Which Caused Creation of Ticket:**

In both cases ACAS threw a generic error message:

Stereo Category: "An error occurred deleting Stereo Category <stereo_category_name>. Please contact support for additional help."
Vendor: "An error occurred deleting Vendor <vendor_name>. Please contact support for additional help."

## How Has This Been Tested?
Manually 

See Examples... 

Delete Stereo Category
Delete Vendor

### DIFF
--- a/modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminBrowser.coffee
+++ b/modules/CodeTablesAdmin/src/client/AbstractCodeTablesAdminBrowser.coffee
@@ -290,10 +290,18 @@ class AbstractCodeTablesAdminBrowserController extends Backbone.View
 				@$(".bv_deletingStatusIndicator").addClass "hide"
 				@$(".bv_codeTablesAdminDeletedSuccessfullyMessage").removeClass "hide"
 				@handleSearchRequested()
-			error: (result) =>
+			error: (response) =>
 				@$(".bv_okayButton").removeClass "hide"
 				@$(".bv_deletingStatusIndicator").addClass "hide"
 				@$(".bv_errorDeletingCodeTablesAdminMessage").removeClass "hide"
+				errorMsg = response.responseText
+				if (errorMsg)
+					errorJSON = JSON.parse(errorMsg)
+					# Grabbing the first (and assumingly only) error and displaying the message in placeholder element 
+					@$('.bv_deleteCodeTablesAdminErrorMessageHolder').html errorJSON[0].message
+				else
+					noServerMessage = "There is no error message from the server."
+					@$('.bv_deleteCodeTablesAdminErrorMessageHolder').html noServerMessage
 		)
 
 	handleCancelDeleteClicked: =>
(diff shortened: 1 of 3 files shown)


dalejerikson:
Questions: 

1. Is is reasonable to assume that models that are used by AbstractCodeTablesAdminBrowserController and have a DELETE route in Roo will have an error message with the JSON structure of [{level: "ERROR", message: <message>}, ...] ? I don't want to break anything else. So far it seems fine. Not sure how much checking I should do of the error message that's returned from the server in the code.

2. Should the error message from the server be appended to the previous message that would be thrown or should it completely replace the original message?

bffrost:
> Is is reasonable to assume that models that are used by AbstractCodeTablesAdminBrowserController and have a DELETE route in Roo will have an error message with the JSON structure of [{level: "ERROR", message: }, ...] ? I don't want to break anything else. So far it seems fine. Not sure how much checking I should do of the error message that's returned from the server in the code.

This is a good point of concern. I think you can check this yourself. You can search the acas codebase for usages of AbstractCodeTablesAdminBrowserController (there are only 4 references), and you can check what Java object each of the associated backends produces, or you can try to test disallowed deletes in all 4 of those modules.

> Should the error message from the server be appended to the previous message that would be thrown or should it completely replace the original message?
It looks a little funny to me, which is I suspect why you're asking this question. I think the bottom error message is clear enough by itself. The only things we need to communicate to the user here are "you can't do this, and here's why", and I think it gets that across.


Some other things that stand out to me:
- We could really use quotes around the specific stereo category or vendor name within that error message. It's a bit hard to parse visually.
- If we can get our hands on _which_ parents or lots are preventing the deletion, that'd be useful info to show the users. We'd likely want to cap the number of parent / lot corp names we show though, to prevent the error message from getting unwieldy in cases of too many dependencies. This is a nice-to-have.

dalejerikson:
Screen Shot 2022-10-17 at 9 13 37 AM
Screen Shot 2022-10-17 at 9 14 14 AM


Changed the messages to be more clear and less redundant

dalejerikson:
Related: https://github.com/mcneilco/acas-roo-server/pull/376

bffrost (approved):
LGTM

PR #1016, opened 2022-10-04
ACAS-466: heart beat monitor in ui

brianbolt:
## Description
Heart beat monitor which checks every 10 seconds for a disconnection or logout of ACAS and displays a message.
If the message is already being shown and the status does not change then the message is left
If the status changes (either a successful connection or not) the message is removed or change to match the current status.

## Related Issue
ACAS-466

## How Has This Been Tested?
Logged into ACAS and restarted the backend and verified a message about not being able to communicate with ACAS showed.
Once ACAS restarted but the user was logged out, verified the message was updated to match new "logged out" status.
Logged back in via another tab and verified that the message went away.

### DIFF
--- a/modules/ModuleMenus/src/client/ModuleMenus.coffee
+++ b/modules/ModuleMenus/src/client/ModuleMenus.coffee
@@ -118,9 +118,12 @@ class ModuleMenusController extends Backbone.View
 						@heartbeatToast = new ACASToast
 							type: "error"
 							title: "Logged out"
-							text: "You have been logged out. Please login again."
+							text: "You have been logged out. You must log back in to continue."
 							position: "bottom-right"
 							duration: 50000
+							onClose: =>
+								# Reload the page if the user clicks the close button on the heartbeat notification
+								location.reload()
 					else if @status != 200
 						@heartbeatToast = new ACASToast
 							type: "error"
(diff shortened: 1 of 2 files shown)


modules/ModuleMenus/src/client/ModuleMenus.coffee
```diff
@@ -88,9 +88,48 @@ class ModuleMenusController extends Backbone.View
...
+				if !active || originalStatus != @status
+					if @heartbeatToast?
+						@heartbeatToast.closeToast()
+						@heartbeatToast.remove()
+					
+					if @status == 401
+						@heartbeatToast = new ACASToast
+							type: "error"
+							title: "Logged out"
+							text: "You have been logged out. Please login again."
```
bffrost, line 121: Can we indicate what the user's supposed to do? Reload the page, open a new tab? Show a link to the login page?
As a user I'd be unclear on how to "Please login again" if I was not being presented with the login page.

bffrost (approved):
LGTM, thanks for those changes in ACASToast.

Added a request for some more instructive messaging, but otherwise LGTM

modules/ModuleMenus/src/client/ModuleMenus.coffee
brianbolt: Thanks, I added a feature to the toast notification widget where you can pass an "onClose" function.  I added a page reload to the heart beat notification for logged out such that when the user initiates a close by clicking the X, the page reloads (this is what LD does, albeit with a different notification than the bacon bar).  I also changed the text of the notification to match LDs "You have been logged out. You must log back in to continue."

bffrost (approved):
LGTM, nice way of doing this.

PR #1017, opened 2022-10-05
ACAS-271: Fallback for when register request is missing originalFileName

brianbolt:
## Description
Fall back to the required fileName key when registering an SDF file

## Related Issue
ACAS-271

## How Has This Been Tested?
Ran acasclient tests.


dalejerikson (approved):
👍

PR #1018, opened 2022-10-06
Fixed .sdf <-> .zip swapping, removed redundant/unused variables

hhan-schrodinger:
## Description

**Problem**
- Users were using CReg Bulk Loader to load .sdf files with .sdf extensions in multiple locations within the file name (artifacts of previous downloads)
- Downloading summary files/current files were broken since the logic to replace ".sdf" with ".zip" only ran once, and was renaming the first ".sdf" in the file string, not the extension at the end. As a result, the link in the panel could not locate the correct file. 

**Fix**
- Removed the functions using .replace(".sdf", ".zip") and .replace(".sdf", {todays_date} + ".sdf") and replaced them with logic that would remove the last four characters of the file name string and then add ".zip" or {todays_date} + ".sdf" 
- Also removed unused variables

## How Has This Been Tested?
1. Renamed a test .sdf file to "compounds_for_bulkloader_1(6).sdf_2022-09-30_validated (3).sdf" 
2. Bulk loaded / registered file.
3. Tried downloading summary, current file, and original file to see if there were incorrect swapping between ".zip" and ".sdf" or if the file could not be found.

### DIFF
--- a/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.html
+++ b/modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.html
@@ -261,7 +261,7 @@
 </script>
 
 <script type="text/template" id="FileRowSummaryView" xmlns="http://www.w3.org/1999/html">
-    <td class='bv_fileName'><a href='<%-fileLink%>'download="<%-fileName%>" ><%-fileName%></a></td>
+    <td class='bv_fileName'><a href='<%-fileLink%>'download="<%-originalFileName%>" ><%-originalFileName%></a></td>
     <td class='bv_loadDate'><%-loadDate%></td>
     <td class='bv_loadUser'><%-loadUser%></td>
     <td class="bv_currentSDF"><a href=<%-currentFileLink%> download="<%-currentFileName%>">Download</a></td>


modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
```diff
@@ -951,21 +950,24 @@ class FileRowSummaryController extends Backbone.View
 			mm = '0' + mm
 		today = '_' + mm + '_' + dd + '_' + yyyy
 
-		# Rename the SDF file representing the current
-		currentFileName = originalFileName.replace "\.sdf", today + "_current_state.sdf"
-    
+		#In some cases, there are artifact .sdf extensions in a file name that cause downloading issues when renanubg a file to .zip
+		#we need to replace ONLY the last .sdf in the file with ".zip" or "{current_date}_current_state.sdf", so we can't use replace which will replace the first
```
hhan-schrodinger, line 954: Should I add logic here to check if the last four characters are actually ".sdf"? I don't think there would be a case where it wouldn't be an .sdf file - but if it is not it would break? I feel like blindly removing the last four characters may not be the best option.

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
```diff
@@ -951,21 +950,24 @@ class FileRowSummaryController extends Backbone.View
...
+		currentFileName = fileNameBase + today + "_current_state.sdf"
+
 		# Replace .sdf with .zip since the report file shares the same name 
-		reportName = fileName.replace "\.sdf", ".zip"
-		reportDisplayName = originalFileName.replace "\.sdf", ".zip"
+		reportName = fileNameBase + ".zip"
 
 		toDisplay =
 			fileName: fileName
-			originalFileName: originalFileName
```
bffrost, line 963: @hhan-schrodinger I'd added this "originalFileName" stuff as part of ACAS-271. Please don't remove it! If this wasn't working on your local environment, you may need to update Roo, since I merged in the changes for ACAS-271 recently.

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
bffrost: I agree this seems a little hacky, but we truly aren't expecting any cases where the file isn't .sdf, as that's the only format we support. I think it's OK as is.

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
```diff
@@ -951,21 +950,24 @@ class FileRowSummaryController extends Backbone.View
...
+		#we need to replace ONLY the last .sdf in the file with ".zip" or "{current_date}_current_state.sdf", so we can't use replace which will replace the first
+		fileNameLength = fileName.length #we have to assign this first since for some reason simply using fileName.length - 4 just wont work in line below? 
+		fileNameBase = fileName.slice 0, (fileNameLength)-4 
+	
+		# Rename the SDF file representing the current data
+		currentFileName = fileNameBase + today + "_current_state.sdf"
+
 		# Replace .sdf with .zip since the report file shares the same name 
-		reportName = fileName.replace "\.sdf", ".zip"
-		reportDisplayName = originalFileName.replace "\.sdf", ".zip"
```
bffrost, line 959: This was important for ACAS-271, where we're trying to preserve the original filename. Can you try applying the same logic to `originalFileName` that you currently do to `fileName`? The idea from ACAS-271 is that we use the originalFileName when generating things the user will see (text in the UI, the downloaded name of the file, i.e. the `download=` attribute in the `a` tag), and use the fileName when generating the paths that need to exist on disk on the ACAS server.

bffrost (changes requested):
The idea of your fix looks right, but you've clobbered my changes from ACAS-271. Please try to restore those.

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
hhan-schrodinger: Ooops, sorry Brian! I assumed it was an artifact/error on my part when I made it.

hhan-schrodinger:
Is removing reportDisplayName okay?

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.coffee
```diff
@@ -951,12 +951,19 @@ class FileRowSummaryController extends Backbone.View
...
-		# Rename the SDF file representing the current
-		currentFileName = originalFileName.replace "\.sdf", today + "_current_state.sdf"
-    
+		#In some cases, there are artifact .sdf extensions in a file name that cause downloading issues when renanubg a file to .zip
+		#we need to replace ONLY the last .sdf in the file with ".zip" or "{current_date}_current_state.sdf", so we can't use replace which will replace the first
+		fileNameLength = fileName.length #we have to assign this first since for some reason simply using fileName.length - 4 just wont work in line below? 
+		fileNameBase = fileName.slice 0, (fileNameLength)-4 
+		#do the same logic for originalFileName
+		originalFileNameLength = originalFileName.length
+		originalFileName =  originalFileName.slice 0, (originalFileName)-4
```
bffrost, line 960: This should be `originalFileNameBase` right? I think you're clobbering `originalFileName` that we still use below.

Once you have `originalFileNameBase` then you can easily assemble the report displayName:
```
reportName = fileNameBase + ".zip"
reportDisplayName = originalFileNameBase + ".zip"
```

modules/CmpdRegBulkLoader/src/client/CmpdRegBulkLoader.html
```diff
@@ -261,7 +261,7 @@ <h4 style="font-size:16px;">Validation Summary:</h4>
 </script>
 
 <script type="text/template" id="FileRowSummaryView" xmlns="http://www.w3.org/1999/html">
-    <td class='bv_fileName'><a href='<%-fileLink%>'download="<%-originalFileName%>" ><%-originalFileName%></a></td>
```
bffrost, line 264: not sure why this change was made. We do want the `download` attribute and the displayed text to be based on the originalFileName, not fileName.

bffrost (changes requested):
@hhan-schrodinger this still isn't fully working with my changes. I tested locally on your branch by bulk loading a file that I'd uploaded loaded several times before.

Screen Shot 2022-10-17 at 2 19 32 PM

It should show `test_005_swap_parent_structures.sdf` and `test_005_swap_parent_structures.zip` in the names of the links, and download as those names, but instead it includes the ` (5)`.

hhan-schrodinger:
@bffrost I made the adjustments, but I'm wondering are we trying to add logic to remove any duplicate file artifacts from the filename ie. "(5)"?

bffrost (approved):
LGTM. @hhan-schrodinger no, we're not doing any explicit logic to try to parse & remove artifacts. Just trying to be faithful about preserving the file name as it was originally uploaded.

This looks ready to merge.

PR #1022, opened 2022-10-12
ACAS-482 fixing registerdBy ACLS permissions

brianbolt:
## Description
Get lot ACLs is currently broken for registeredBy users.  Instead of checking for registedBy, we are checking for a non-existent recordedBy attribute.  This causes users who registered a lot but aren't cmpdreg admins, or the "chemist" of the lot to be able to edit their own lots.

## Related Issue
ACAS-482

## How Has This Been Tested?
Configured a lot and permissions such that I reproduced the bug and could not edit a lot that was registeredBy me.
Updated the code and refreshed the page and verified I could not edit the lot.
Switched the registedBy user to another user and verified I could no longer edit the lot. (negative case test).


dalejerikson (approved):
👍

PR #1026, opened 2022-10-19
ACAS-488: Allow user to limit by max results input

brianbolt:
## Description
 - Allow the user to limit the download sdf results by passing the max results input to the service

## Related Issue
ACAS-488

## How Has This Been Tested?
Clicked download and verified the results were limited
Removed the input and verified that the results were not limited by max results


bffrost (approved):
LGTM

PR #1027, opened 2022-10-20
ACAS-486: Standardization dry run UI change to go with service changes

brianbolt:
## Description
 - Changes to the UI to match the service changes.
    -  Parent is now added to the response json with corpName and molWeight as attributes instead of being top level
 - Use delta mol weight calculated by the server to keep logic in one place

Changes. to be merged in only when closing ACAS-480-epic branch.

## Related Issue
ACAS-486

## How Has This Been Tested?
Ran queries and made sure UI looked correct for delta mol weight, old mol weight and corp name fields and that structures rendered in the UI (as they use the corp name returned to fetch images).


bffrost (approved):
LGTM

PR #1029, opened 2022-11-01
Fix backwards compability with older maestro sketcher versions which used sketcherExportMolBlock

brianbolt:
## Description
  - MaestroJSUtil was missing a check for the older `sketcherExportMolBlock` to be backwards compatible with older maestro sketchers

## Related Issue
ACAS-469

## How Has This Been Tested?
Updated a server running older maestro sketcher and verified the sketcher loaded.

### DIFF
--- a/modules/CmpdReg/src/client/schrodinger/maestrosketcher/maestrosketcherlauncher.js
+++ b/modules/CmpdReg/src/client/schrodinger/maestrosketcher/maestrosketcherlauncher.js
@@ -6,7 +6,7 @@ $(function() {
             function waitForReady (resolve, reject) {
                 var _a;
                 maestro = $(el)[0].contentWindow.Module;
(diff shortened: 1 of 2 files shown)


bffrost (commented):
@brianbolt based on ACAS-205 I think we may also need to make a similar backwards compatibility fix on `setSketcherMolBlock` / `sketcherImportText`

Can you check if that's the case or not?

bffrost (approved):
@brianbolt as long as you've tested pasting a structure into the sketcher on your test instance, LGTM

brianbolt:
Updated to test for sketcher readiness by testing for `sketcherImportText` rather than `sketcherExportText` to be consistent
The ACASFormChemicalStructure.coffee file should have used `sketcherImportText` rather than the really old `sketcherImportMolBlock`

I tested these changes against LD 2022-2

PR #1030, opened 2022-11-03
Modify SEL to check endpoints against protocol rather than value_kind

hhan-schrodinger:
## Description
Added features to generic_data_parser.R script to work with the new endpoint manager when experiments are loaded. 

When an experiment is loaded that will create a new protocol, added logic that will: 
- Save the same "data column order" information currently being saved to Experiment via "createColumnOrderStates" into the new Protocol created by "createNewProtocol"
- Save incoming column names into the "data column"/"column name" codetable / data dictionary
- Create a value for "Strict Endpoint Matching" when a new protocol is created via experiment (based on config file)

When an experiment is loaded against an existing protocol that has strict endpoint matching enabled, added logic that will: 
- Instead of validating column headers against the "value_kind" table, validate them against the "data column order" lsStates of the protocol.
- Implement the warning & error messages that indicate the experiment columns that do not match the existing protocol's endpoints

## How Has This Been Tested?
- Loaded experiments to create new protocols and saw that the new column names/units/data types have been saved as new endpoints in the endpoint manager
- Loaded experiments against existing protocols with strict endpoint matching enabled to test that they threw errors indicating when the experiment column did not match the protocol's endpoints, and that they passed when they did match

Example of Error: 
Screen Shot 2022-11-04 at 12 07 07 PM

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -3512,14 +3512,12 @@ runMain <- function(pathToGenericDataFormatExcelFile, reportFilePath=NULL,
 
   # Validate Experiment Columns against Protocol Endpoints
   # only proceed if it associated with an existing protocol (the protocol is not new)
-  if (!newProtocol) {
-    #if the protocol has strict endpoint matching enabled, proceed with validation
-    if (getProtocolStrictEndpointMatching(protocol) == TRUE) {
-      #extract the endpoint data from the protocol object to check against
-      protocolEndpointData <- getProtocolEndpointData(protocol)
-      #Check the experiment columns against the protocolEndpointData
-      validateExperimentColumns(selColumnOrderInfo, protocolEndpointData)
-    } 
+  # also, only proceed if endpoint manager is enabled
+  if (!newProtocol && racas::applicationSettings$client.protocol.endpointManager.enabled) {
+    #extract the endpoint data from the protocol object to check against
+    protocolEndpointData <- getProtocolEndpointData(protocol)
+    #Check the experiment columns against the protocolEndpointData
+    validateExperimentColumns(selColumnOrderInfo, protocolEndpointData, getProtocolStrictEndpointMatching(protocol))
   }
   
   # If there are errors, do not allow an upload
@@ -4061,13 +4059,16 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
 }
 
 getProtocolEndpointData <- function(protocol) {
+  library(dplyr)
   # Collects all the endpoint values in the protocol object and returns a data frame of them
 
   # First, we create an empty dataframe for the endpoint data associated with the protocol 
   protocolEndpointDataFrame <- data.frame(
     columnName=character(),
     units=character(),
-    dataType=character()
+    dataType=character(),
+    conc = character(),
+    concUnits = character()
   )
 
   #go through the protocol data to wrangle the endpoint data into the dataframe
@@ -4108,12 +4109,12 @@ getProtocolEndpointData <- function(protocol) {
         dataType = dataTypeEntry, 
         conc = concEntry, 
         concUnits = concUnitsEntry)
-
-      protocolEndpointDataFrame = rbind(protocolEndpointDataFrame, dataEntry, stringsAsFactors=FALSE)
-
+      
+      
+      protocolEndpointDataFrame = bind_rows(protocolEndpointDataFrame, dataEntry)
     }
   }
-
+  
   return(protocolEndpointDataFrame)
 }
 
@@ -4138,7 +4139,7 @@ checkIfEndpointsMatch <- function(protocolValue, experimentValue) {
   }
 }
 
-validateExperimentColumns <- function(selColumnOrderInfo, protocolEndpointDataFrame) {
+validateExperimentColumns <- function(selColumnOrderInfo, protocolEndpointDataFrame, protocolStrictEndpointMatchingEnabled) {
   # Checks if the experiment columns names/units/data type are found in the protocol endpoint data
   # Raises errors if an experiment column is not valid 
   # Only used for protocols with strict endpoint enabled
@@ -4177,10 +4178,15 @@ validateExperimentColumns <- function(selColumnOrderInfo, protocolEndpointDataFr
     }
 
     if (experimentRowMatchesEndpoint == FALSE) {
-      addError(paste0("The result type '", experimentRowName, "' with data type '", experimentRowDataType, "' and units '", experimentRowUnits, "' is not one of the allowed result types for this protocol. Please revise your file or contact an ACAS administrator to update the allowed result types for this ", racas::applicationSettings$client.protol.label, "."))
+      # If strict endpoint matching is enabled, we throw an error, otherwise we throw a warning
+      if (protocolStrictEndpointMatchingEnabled == TRUE) {
+        addError(paste0("The result type '", experimentRowName, "' with data type '", experimentRowDataType, "' and units '", experimentRowUnits, "' is not one of the allowed result types for this ", racas::applicationSettings$client.protocol.label, ". Please revise your file or contact an ACAS administrator to update the allowed result types for this ", racas::applicationSettings$client.protocol.label, "."))
+      } else if (protocolStrictEndpointMatchingEnabled == FALSE) {
+        warnUser(paste0("The result type '", experimentRowName, "' with data type '", experimentRowDataType, "' and units '", experimentRowUnits, "' is not one of the defined result types for this ", racas::applicationSettings$client.protocol.label, ". You may want to revise your file to update the allowed result types for this ", racas::applicationSettings$client.protocol.label, "."))
+      } 
     }
+      
   }
-
 }
 
 getProtocolStrictEndpointMatching <- function(protocol) {
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -1901,7 +1901,7 @@ getPreferredProtocolName <- function(protocol, protocolName = NULL) {
   }
   return(preferredName)
 }
-createNewProtocol <- function(metaData, lsTransaction, recordedBy) {
+createNewProtocol <- function(metaData, lsTransaction, recordedBy, columnOrderStates) {
   # creates a protocol with the protocol name and scientist in the metaData
   # 
   # Args:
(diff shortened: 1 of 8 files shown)


modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -223,6 +223,77 @@ validateMetaData <- function(metaData, configList, username, formatSettings = li
 
   return(list(validatedMetaData=validatedMetaData, duplicateExperimentNamesAllowed=duplicateExperimentNamesAllowed, useExisting=useExisting))
 }
+
+
+saveIncomingEndpointData <- function(codes, codeKind) {
```
bffrost, line 228: nit: this function name is pretty broad. Something narrower like `saveEndpointDDictValues` or `saveEndpointCodeTables` might be clearer as to what's in and out of scope of this function.

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -223,6 +223,77 @@ validateMetaData <- function(metaData, configList, username, formatSettings = li
...
+  # Saves incoming column data into their codetable / data dictionaries
+  # set up for endpoint manager data, only looking at "column name", "column units", "concentration units", "time units" for now... 
+  # We are going off just starting to just take one input columnName to save
```
bffrost, line 231: nit: Is this comment line still relevant?

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -223,6 +223,77 @@ validateMetaData <- function(metaData, configList, username, formatSettings = li
...
+  ddictValues <- rbindlist(lapply(ddictLists, jsonlite::fromJSON))
+
+  # if the object is empty, we assume that no codes exist yet, so we will save all the input codes
+  if(nrow(ddictValues) == 0) {
+    codesToSave <- codes
+
+  #if the object is not empty, we will check to see which of the existing codes overlap with our input codes
+  } else {
+    #extract all the codes for a given codeKind
+    codesForCodeKind = ddictValues %>% filter(lsKind == codeKind) %>% pull(codeName)
```
bffrost, line 265: I'm not sure `codeName` is the right attribute to pull here. If I'm understanding the flow so far,
codeName will be something like 'DDICT-00001" whereas the `shortName` attribute will be the relevant string containing the column name or units. Could you double check this?

You'll also be able to test this by uploading the same column name in two experiments, and seeing if you get duplicate entries in the Column Name dropdown. You can also confirm in the `ddict_value` table

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3421,6 +3510,18 @@ runMain <- function(pathToGenericDataFormatExcelFile, reportFilePath=NULL,
       addError(paste0("Caught an error trying to validate dose response data. Please contact your administrator.: ", e$message))
     })
   }
+
+  # Validate Experiment Columns against Protocol Endpoints
+  # only proceed if it associated with an existing protocol (the protocol is not new)
+  if (!newProtocol) {
+    #if the protocol has strict endpoint matching enabled, proceed with validation
+    if (getProtocolStrictEndpointMatching(protocol) == TRUE) {
```
bffrost, line 3518: We should do this validation either way. But the "strict" endpoint matching value should determine whether the validation throws errors (if strict) or warnings (if not strict).

It might make sense to pass the strictEndpointMatching value into `validateExperimentColumns` as an input variable.

However I think it would be a good idea to check the `client.protocol.endpointManager.enabled` config and skip this whole validation if that is set to false.

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3442,9 +3543,20 @@ runMain <- function(pathToGenericDataFormatExcelFile, reportFilePath=NULL,
...
   }
   
+  # Save endpoint data codes
+  saveIncomingEndpointData(selColumnOrderInfo$Units, "column units")
+  saveIncomingEndpointData(selColumnOrderInfo$valueKind, "column name")
+  saveIncomingEndpointData(selColumnOrderInfo$concUnits, "concentration units")
+  saveIncomingEndpointData(selColumnOrderInfo$timeUnit, "time units")
+
+  ## SEL column order info
+  #saveSession('insideselColumnOrderInfo.RData')
```
bffrost, line 3553: nit: you can remove this now since you're done debugging

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3953,6 +4063,133 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
...
+          } else {
+            return(FALSE)
+          }   
+        }
+      }
+    }
+  }
+
+  # TODO - remove this after database migration
+  return(FALSE)
```
bffrost, line 4190: I think it'd make more sense to return `racas::applicationSettings$client.protocol.strictEndpointMatchingDefault` if the protocol doesn't have this value set. It is the default after all.

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3953,6 +4063,133 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
...
+      #if the experiment column matches one of the endpoints for the protocol, record it and move onto the next column
+      if (rowNamesMatch & rowUnitsMatch & rowTypesMatch) {
+        experimentRowMatchesEndpoint = TRUE
+        break 
+      }
+
+    }
+
+    if (experimentRowMatchesEndpoint == FALSE) {
+      addError(paste0("The result type '", experimentRowName, "' with data type '", experimentRowDataType, "' and units '", experimentRowUnits, "' is not one of the allowed result types for this protocol. Please revise your file or contact an ACAS administrator to update the allowed result types for this protocol."))
```
bffrost, line 4162: Here's where you'll want to split between error & warning based on the "strict"ness of the protocol.

See the last image on https://confluence.schrodinger.com/x/egSgBw for some proposed verbiage.

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3953,6 +4063,133 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
...
+  # Checks if the experiment columns names/units/data type are found in the protocol endpoint data
+  # Raises errors if an experiment column is not valid 
+  # Only used for protocols with strict endpoint enabled
+
+  for (experimentRowNum in seq(1, nrow(selColumnOrderInfo))) {
+    experimentRowData <- selColumnOrderInfo[experimentRowNum,]
+    experimentRowName = experimentRowData$valueKind
+    experimentRowUnits = experimentRowData$Units
+    experimentRowDataType = experimentRowData$valueType
+
```
bffrost, line 4121: We should check concentration and conc units too

bffrost (changes requested):
@hhan-schrodinger this is in pretty good shape. The code is well structured and easy to follow, especially with your code comments. I found a few areas that differ from the design but they all look like easy fixes.

dalejerikson (commented):
I can't really comment in the context of the user story. However, I do really like your styling and comments. I think it's really clean and me actually not mind reading R (which is hard to do haha).

modules/GenericDataParser/src/server/generic_data_parser.R
hhan-schrodinger: Makes sense!

modules/GenericDataParser/src/server/generic_data_parser.R
hhan-schrodinger: Good catch!

modules/GenericDataParser/src/server/generic_data_parser.R
hhan-schrodinger: Should we have additional logic to only validate if conc/conc units are enabled in the conf?

bffrost: That's a great question. Yes, we should add some logic for that case. ACAS is configured to not show concentration in the UI, and someone attempts to upload data that has concentrations into a strict match protocol, we should at the very least warn them. Something along the lines of:

```
The column you're uploading {column name} at concentration {conc} {conc units} cannot be fully validated against protocol {protocol name} because this ACAS server is not configured to track concentration in the Protocol Endpoints. To enable complete checking including concentration, please contact your administrator and ask them to enable "client.protocol.endpointManager.showConcentration"
```

modules/GenericDataParser/src/server/generic_data_parser.R
hhan-schrodinger: You are completely right! I think I misread the table. 

Screen Shot 2022-11-16 at 12 58 39 PM

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3953,6 +4063,133 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
...
+}
+
+getProtocolStrictEndpointMatching <- function(protocol) {
+  # Check whether the protocol has strict endpoint matching set or not 
+  # Returns true or false
+
+  # TODO - This may cause errors in existing protocols that do not have strict endpoint matching already set
+  # To prevent those errors, we will automatically return false for now. 
+
+  for (lsState in protocol$lsStates) {
```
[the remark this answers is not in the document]
hhan-schrodinger: Thanks, thats super helpful!

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -4064,13 +4058,16 @@ getFormatParameters <- function(rawOnlyFormat, customFormatSettings, inputFormat
...
   # Collects all the endpoint values in the protocol object and returns a data frame of them
 
   # First, we create an empty dataframe for the endpoint data associated with the protocol 
   protocolEndpointDataFrame <- data.frame(
     columnName=character(),
     units=character(),
-    dataType=character()
+    dataType=character(),
+    conc = character(),
+    concUnits = character()
```
bffrost, line 4070: Are we missing Time & Time Units here?

bffrost (approved):
LGTM. I think I see a few places where we're not validating Time, but I'm comfortable with spinning that off as a separate case so you can get this PR merged. Nice work!

modules/GenericDataParser/src/server/generic_data_parser.R
hhan-schrodinger: I meant to ask - I wasn't sure whether or not to include Time/Time Units initially. Can definitely add them here and the rest of the logic!

PR #1031, opened 2022-11-03
ACAS-498: Add Writeup field to CReg analytical files

bffrost:
## Description
ACAS-498 adds a new "Writeup" text field associated with each analytical file attached to a Lot in CReg. The purpose of this is to store standard NMR and LCMS writeup information (https://pubsapp.acs.org/paragonplus/submission/acs_nmr_guidelines.pdf) used in publications, alongside the corresponding pdfs and raw data.

The frontend portion here adds the new `textarea` to the form, makes sure it is saved and rendered properly, and tweaks some styling so that saved and new rows line up properly. See screenshot below.

## How Has This Been Tested?
Tested locally that I can upload and save an analytical file with a writeup. Confirmed that it's rendered properly when the page is reloaded after save.
Screen Shot 2022-11-03 at 1 05 37 PM

### DIFF
--- a/modules/CmpdReg/src/client/MultipleFilePicker/css/style_nocompile.css
+++ b/modules/CmpdReg/src/client/MultipleFilePicker/css/style_nocompile.css
@@ -9,7 +9,7 @@ body {
 
 div#upload-app {
     /* ORIGINAL WIDTH width: 510px;*/
-    width: 625px;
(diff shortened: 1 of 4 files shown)


brianbolt (approved):
LGTM

dalejerikson (approved):
👍

PR #1032, opened 2022-11-03
ACAS-491: GitHub actions fix deprecation warnings

bffrost:
## Description
Fixing warnings:
- Bump `docker/*`, `setup-python`, and `github-script` action versions to fix Node 12 deprecation warnings
- Replaced set-output based on https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

A few small unrelated changes:
- Applied a fix for sorting git tags (`--sort=v:refname` rather than `--sort=committerdate`) that I'd applied to acas-roo-server and racas but forgot to apply to acas
- Removed `.github` from gitignore because we do want to track the workflow files in there

## How Has This Been Tested?
Tested via this PR.
Latest run has no warnings: https://github.com/mcneilco/acas/actions/runs/3389274203


brianbolt (approved):
LGTM

Issue #394, opened 2017-11-13
Update package.json for 1.12.3


samanthamunoz:
Update all versions with ">" to tested PROD versions.

PR #1033, opened 2022-11-04
Switch "data column order" stringValues to codeValues

hhan-schrodinger:
## Description
For the endpoint manager, the items in the "data column order " state have been saved with "stringValue" types when they should be saved with "codeValue" types to follow ACAS's standards.

Changed the following values from "stringValue" to "codeValue", and set the "codeType", "codeKind", and "codeOrigin" attributes accordingly in Frontend Protocol/Experiment endpoint tables and Backend R logic (genericDataParser):

- column name
- column conc units
- column time units
- column units
- column type
- hide column
- condition column

## How Has This Been Tested?
- Used experiment loader to upload experiment to see if the experiment and new protocol values were stored as codeValues and if the values were displayed in the endpoint table
- Also tested by manually creating a new protocol

### DIFF
--- a/modules/Components/src/client/ACASFormFields.coffee
+++ b/modules/Components/src/client/ACASFormFields.coffee
@@ -1034,7 +1034,7 @@ class ACASFormLSBooleanFieldController extends ACASFormAbstractFieldController
 
 	renderModelContent: =>
 		# If value is anything other than true (i.e. null), then default to unchecked
-		if @getModel().get('value') is "true"
+		if @getModel().get('value').toLowerCase() is "true"
 			@$('input').attr 'checked', 'checked'
 		else
 			@$('input').removeAttr 'checked'
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -4088,11 +4088,11 @@ getProtocolEndpointData <- function(protocol) {
 
           #depending on what the lsKind is, record the data
           if (value[['lsKind']] == "column name") {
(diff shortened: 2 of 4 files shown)


bffrost:
@hhan-schrodinger I noticed a bug in my local testing where "hide column" values created in R had codeValue "TRUE" / "FALSE" but the frontend was rendering them all as an unchecked box.
I've added a fix for that into your branch, so it's now part of this PR.

bffrost (approved):
LGTM

dalejerikson:
👍

dalejerikson (approved):
👍

modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -1034,7 +1034,7 @@ class ACASFormLSBooleanFieldController extends ACASFormAbstractFieldController
 
 	renderModelContent: =>
 		# If value is anything other than true (i.e. null), then default to unchecked
-		if @getModel().get('value') is "true"
+		if @getModel().get('value').toLowerCase() is "true"
```
brianbolt, line 1037: I just commented about this on [#394](https://github.com/mcneilco/acas-roo-server/pull/394) about how we are storing upper case true values in the migration.

Also, did you check that this also work in the case where the value is null?

hhan-schrodinger: Looked at it, not sure if I fully understand...whether it is stored as "TRUE"/"FALSE" or "true"/"false", this should be OK right? 

Will look into what happens if it is null.

PR #1034, opened 2022-11-09
Bug Fix: Duplicate Parent Alias Errors Don't Bubble Up To UI

dalejerikson:
**Steps to Reproduce:**

With the setting to disallow duplicate aliases turned on. Try and edit a parent alias to match a parent alias that exists on another compound already.

**New Results:**

The UI displays an error to the user stating that:

Parent Aliases must be globally unique.

**Previous Results:**

No error is shown and the UI hangs

## How Has This Been Tested?
Manually checking steps provided in ticket. See Screenshot... 

Screen Shot 2022-11-09 at 11 42 15 AM

### DIFF
--- a/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
+++ b/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
@@ -600,7 +600,9 @@ exports.saveMetaLot = (req, resp) ->
 	, (error, response, json) =>
 		if !error
 			console.log JSON.stringify json
-			if JSON.stringify(json).indexOf("Duplicate") != -1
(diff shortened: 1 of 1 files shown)


dalejerikson:
@brianbolt Not sure if there's a cleaner fix than what I provided or if it's what you had in mind. Let me know what you think.

brianbolt (approved):
Yeah, thanks for pointing that out. I think a cleaner fix would be to actually set the status code from roo to be 409 instead of testing for a specific string but I think this is fine for now especially since I'm the one who originally took the shortcut here.

LGTM!

bffrost (approved):
LGTM

hhan-schrodinger (approved):
Makes sense! Did not know how finding the error in the JSON and reporting it worked before!

PR #1035, opened 2022-11-09
Fix: ACAS gives confusing error when user does not have project access to protocol

dalejerikson:
**Steps to Reproduce**

As an ACAS user, attempt to upload an experiment to a protocol which is restricted to project you do not have access to.

**Fix Results**

ACAS produces a well-formatted, readable error message that is intelligible to the user.

("The [protocol name] you entered is being used in a project that you do not have access to.") 

**Previous Buggy Results**

ACAS produces the following message:
"unused arguments (racas::applicationSettings$client.protocol.label, " you entered is being used in a project that you do not have access to.")


brianbolt (approved):
LGTM

bffrost (approved):
LGTM. @dalejerikson can you look at adding an acasclient test for this? I see we do have some tests creating restricted projects and testing project restrictions on lots, but we don't seem to have any on protocols & experiments. This one is a good targeted test case and we can confirm we get the expected error message.

hhan-schrodinger (approved):
👍

PR #1037, opened 2022-11-10
Bug Fix: Open in Livedesign Button Broken in Curve Curator

dalejerikson:
**Extra Context** 

The efforts to DRY the "Open In" button feature with 
OpenExperimentInQueryTool was not applied to the button in Curve Curator. It was added to the Experiment Browser and the Experiment Loader but forgotten here. 

**Steps to Reproduce:**

Load a Dose Response formatted file into ACAS using the Dose Response module

Open the data set in Curve Curator

Click the Open In LiveDesign button

**Previous Bug Results:**

Browser opens a new tab but Live Design never opens

**Fix Results:**

LiveDesign opens to a Live Report with the expected data

### DIFF
--- a/modules/CurveAnalysis/src/client/CurveCuratorAppController.coffee
+++ b/modules/CurveAnalysis/src/client/CurveCuratorAppController.coffee
@@ -37,11 +37,11 @@ class CurveCuratorAppController extends Backbone.View
 	loadCurvesForExptCode: (exptCode, curveID) =>
 		UtilityFunctions::showProgressModal @$('.bv_loadCurvesModal')
 		@ccc.setupCurator(exptCode, curveID)
-		resultViewerURL = "/openExptInQueryTool?experiment="+exptCode
-		@$('.bv_resultViewerBtn').attr('href',resultViewerURL)
-		@$('.bv_resultViewerBtn').html('Open in '+window.conf.service.result.viewer.displayName)
-		@$('.bv_resultViewerBtn').show()
 
+		@openExperimentInQueryToolController = new OpenExperimentInQueryToolController
+			code: exptCode 
+			experimentStatus: 'approved'
+		$('.bv_openExperimentInQueryToolPlaceholder').html @openExperimentInQueryToolController.render().el
 
 	hideLoadCurvesModal: =>
 		UtilityFunctions::hideProgressModal @$('.bv_loadCurvesModal')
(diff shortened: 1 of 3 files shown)


dalejerikson:
Screen Shot 2022-11-10 at 1 23 53 PM
Curve Curator now uses "Open Experiment In Query Tool" class (Introduced in: https://github.com/mcneilco/acas/pull/964/files) as opposed to a previous deprecated button to Open in Live Design.

modules/CurveAnalysis/src/client/CurveCuratorAppController.coffee
```diff
@@ -37,11 +37,11 @@ class CurveCuratorAppController extends Backbone.View
...
 		UtilityFunctions::showProgressModal @$('.bv_loadCurvesModal')
 		@ccc.setupCurator(exptCode, curveID)
-		resultViewerURL = "/openExptInQueryTool?experiment="+exptCode
-		@$('.bv_resultViewerBtn').attr('href',resultViewerURL)
-		@$('.bv_resultViewerBtn').html('Open in '+window.conf.service.result.viewer.displayName)
-		@$('.bv_resultViewerBtn').show()
 
+		@openExperimentInQueryToolController = new OpenExperimentInQueryToolController
+			code: exptCode 
+			experimentStatus: 'approved'
```
brianbolt, line 43: Why is `approved` hardcoded here?  It seems like `OpenExperimentInQueryToolController` should be able look up the experiment status if not provided, or we should pass the real experiment status here.

brianbolt (changes requested):
I don't think we should hardcode `approved` here.

bffrost (approved):
LGTM

modules/CurveAnalysis/src/client/CurveCuratorAppController.coffee
bffrost: @brianbolt this is valid feedback, but it looks like we have 'approved' hardcoded in several other places that use this controller. I'm inclined to merge this as-is, then spin off a separate techdebt case to address the hardcoded experiment statuses.

brianbolt: I see. Makes sense. LGTM then.

brianbolt (approved):

modules/CurveAnalysis/src/client/CurveCuratorAppController.coffee
dalejerikson: Resolving conversation to merge.

PR #1040, opened 2022-11-17
Endpoints should not be added during failed uploads to ACAS

hhan-schrodinger:
## Description

**Expected Results:**
ACAS errors and the endpoints never show up in the endpoints dropdown list because the experiment was never successfully loaded

**Actual Results:**
ACAS errors as expected, however the endpoints are listed in the endpoints dropdown

**Fix:**
- Added logic to the validateExperimentColumns() function so that it returns TRUE/FALSE if the experiment did not pass strict endpoint matching requirement. 
- Added logic so endpoints are only saved if the experiment is creating a new protocol or passes the strict endpoint matching requirement 

## How Has This Been Tested?
Tested with the files used in reporting the bug (file-1.csv, file-2.csv). 

**Before ("Clint Eastwood" endpoint saved when it should not be)**
Screen Shot 2022-11-16 at 7 23 49 PM

**After (No "Clint Eastwood" endpoint saved)**
Screen Shot 2022-11-16 at 7 22 18 PM

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -3544,22 +3544,18 @@ runMain <- function(pathToGenericDataFormatExcelFile, reportFilePath=NULL,
     customExperimentMetaDataValues <- NULL
   }
   
-  # Save endpoint data codes if it passed validation OR if it is a new protocol
-  # (columns don't conflict with protocol endpoints when strict endpoint matching is enabled)
-  if (experimentPassedEndpointValidation == TRUE | newProtocol == TRUE) {
-    saveEndpointCodeTables(selColumnOrderInfo$Units, "column units")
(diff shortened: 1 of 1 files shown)


bffrost (commented):
The code changes make sense to me, but I'm wondering if this covers a broad enough set of cases. There are other reasons an upload could fail, besides the column validation (such as due to bugs in ACAS that cause internal errors). Is there another place lower down in the `runMain` function where we have info about whether the data load was successful or not?
If so we could maybe shift the `saveEndpointCodeTables` logic down there, and have it be one of the last things in `runMain`.

hhan-schrodinger:
@bffrost Makes sense! I moved the saving code logic to the section that checks if there have been any errors, if the protocol is new, and if it is not a dry run. Should be more robust now?

bffrost (approved):
LGTM, I'm much more comfortable with the new spot for those lines

PR #1042, opened 2022-11-18
Textbox alignment in endpoint manager

hhan-schrodinger:
## Description

Fixed fields in the endpoint manager that were not aligned. 
Screen Shot 2022-11-18 at 12 10 24 PM

## How Has This Been Tested?
Tested on endpoint table to see if fields were aligned. 

After:
Screen Shot 2022-11-18 at 12 03 54 PM


bffrost (approved):
LGTM

PR #1044, opened 2022-11-18
Fixed internal error when uploading some experiment files to ACAS

hhan-schrodinger:
## Description
ACAS was giving the following error when uploading some experiment files. 
Screen Shot 2022-11-18 at 12 13 47 PM

The error occurred in experiment files empty values for units that could not be saved when the codes were saved to the database. The issue was resolved by adding handling for empty values when saving codes. 

Also fixed an issue where some objects would not render since they had a null value. 

## How Has This Been Tested?
Reloaded the same files that were previously throwing the error and saved them successfully. Uploaded multiple times by editing protocol to include and not include the endpoints, and tried with strict endpoint enabled and disabled to make sure strict endpoint matching worked correctly. 

Screen Shot 2022-11-18 at 12 23 45 PM


bffrost (approved):
LGTM

PR #1047, opened 2022-11-29
ACAS-507: Lock down edit parent and swap structures endpoints to CReg Admins

bffrost:
## Description
ACAS-507 is meant to open the "Edit Parent" workflow up via acasclient. While doing so, I realized that the two routes used in edit parent, as well as the swapParentStructures endpoints are accessible to any CReg user. Both Edit Parent and Swap Structures should only be accessible to CReg admins. This PR makes this the case.

## How Has This Been Tested?
ACASClient testing in this PR: https://github.com/mcneilco/acasclient/pull/104


brianbolt (approved):
LGTM

PR #1050, opened 2022-12-07
Add button to Protocol Browser that downloads a template SEL file

hhan-schrodinger:
## Description
Created a button for downloading a template SEL file from a given protocol and its endpoints. 

Main Additions:
- Created an API route that creates a template SEL file for a given protocol (/api/getTemplateSELFile) 
- Added UI element to generate and download template SEL file from Protocol editor and Protocol Browser

Other:
- Refactored endpoint management code to only pull data related experiments associated with the protocol once, and re-use it instead of making multiple calls  
- Fixed bug where when a row was clicked on in the endpoint manager, the filtering of associated experiments with that endpoint wasn't working properly. Caused by using stringValue instead of codeValue to check data. 

## How Has This Been Tested?
Examples of download template button in Protocol Editor and Browser and and template SEL file from pressing them. 
Screen Shot 2022-12-07 at 10 07 48 AM
Screen Shot 2022-12-07 at 10 07 24 AM
Screen Shot 2022-12-07 at 10 08 33 AM


modules/ProtocolBrowser/src/client/ProtocolBrowser.coffee
```diff
@@ -442,6 +443,140 @@ class ProtocolBrowserController extends Backbone.View
 		$(".bv_protocolBaseControllerContainer").addClass("hide")
 		$(".bv_noMatchesFoundMessage").addClass("hide")
 
+	downloadSELFile: =>
+		# To generate the SEL file, we need the endpoint data from a different protocol data object we make a request for here: 
+		$.ajax
+			type: 'GET'
+			url: "/api/protocols/codename/" + @protocolController.model.escape('codeName')
```
bffrost, line 450: Just to confirm, we didn't already have the "data column order" states in @protocolController.model ?

modules/ServerAPI/src/server/routes/ProtocolRoutes.coffee
```diff
@@ -0,0 +1,97 @@
...
+
+        endpointNameRowString = "Corporate Batch ID,"
+        dataTypeRowString = "Datatype,"
+        for indexNum in [0..endpointNames.length]
+            endpointRowEntry = ""
+            dataTypeEntry = ""
+
+            endpointHasNoValues = true
+
+            if endpointNames[indexNum] not in blankElements
```
bffrost, line 35: nit: IMO it'd be more readable if you just set the endpointName, unit, conc, etc. to variables up front:
```
endpointName = endpointNames[indexNum]
endpointUnit = endpointUnits[indexNum]
endpointConc[indexNum]
...
```

bffrost (changes requested):
I think you should refactor some of the attribute lookups to the server side (routes). This will simplify the frontend and get rid of duplicated code. It'll also mean shuttling less data back & forth between the frontend and backend, and it'll set us up much better to expose `getTemplateSELFile` as an acasclient API in the future.

modules/ProtocolBrowser/src/client/ProtocolBrowser.coffee
hhan-schrodinger: We do have the "data column order" states, but we don't have the protocol project information in the @protocolController.model as far as I can see which we need.

hhan-schrodinger: No longer needed, logic moved to backend

modules/ServerAPI/src/server/routes/ProtocolRoutes.coffee
hhan-schrodinger: Done!

modules/ProtocolBrowser/src/client/ProtocolBrowser.coffee
```diff
@@ -442,6 +443,34 @@ class ProtocolBrowserController extends Backbone.View
...
+				# exporting and downloading the file to the user
+				encodedUri = encodeURI(response)
+				a = document.createElement('a');
+				a.style.display = 'none';
+				a.href = encodedUri;
+				a.setAttribute('target', '_blank')
+				document.body.appendChild(a);
+				a.click();
+			error: (err) =>
+				console.log "getTemplateSELFile() error:" + err
```
brianbolt, line 470: Can you add something for the user to know that their download failed including the err message?

modules/ServerAPI/src/client/Protocol.coffee
```diff
@@ -802,21 +803,33 @@ class EndpointListController extends AbstractFormController
...
 		$.ajax
 			type: 'GET'
 			#there are two similar routes in ExperimentBrowserRoutes.coffee and ExperimentServiceRoutes.coffee
 			#url: "/api/experimentsForProtocol/#{protocolCode}" #ExperimentBrowserRoutes.coffee route
 			url: "/api/experiments/protocolCodename/#{protocolCode}" #ExperimentServiceRoutes.coffee route
 			success: (experiments) =>
-				@setupExperimentSummaryTable experiments
-				@$(".bv_experimentTableControllerTitle").html "Experiments using " + protocolCode + ":"
+				#save the experiments so we don't have to retrieve them again 
+				@protocolExperiments = experiments
```
brianbolt, line 817: experiments here is an array of javascript objects. It would simplify a lot of your loops through the states and values later on in the code to turn these javascript objects into a Backbone collection of Experiment Backbone objects.

You can change this line to:
```
@protocolExperiments = new ExperimentList experiments
```

Then later on in the code like on line 906 you can use `@protocolExperiments.filter` with experiment methods like `dataColumnOrderState = experiment.get("lsStates").getStatesByTypeAndKind "metadata", "data column order"`

And if the experiment has that state, you can then call:
```
dataColumnOrderState.getValuesByTypeAndKind("column name")
```

modules/ServerAPI/src/client/Protocol.coffee
```diff
@@ -1120,3 +1127,110 @@ class EndpointListController extends AbstractFormController
...
+		endpointNames = []
+		endpointUnits = []
+		endpointDataTypes = []
+		endpointConc = []
+		endpointConcUnits = []
+		endpointTime = []
+		endpointTimeUnits = []
+		endpointHidden = []
+
+		for lsState in @collection
```
brianbolt, line 1143: This would all greatly benefit from using Backbone collection methods and State class methods:

https://github.com/mcneilco/acas/blob/7eba2bfe3ff6ec0aa039ab775cc0b045857b6ef1/modules/ServerAPI/src/client/Label.coffee#L201

modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
```diff
@@ -616,3 +618,197 @@ exports.bulkPutProtocolsInternal = (protsArray, callback) ->
...
+						endpointDataTypes = []
+						endpointConcs = []
+						endpointConcUnits = []
+						endpointTimes = []
+						endpointTimeUnits = []
+						endpointHiddens = []
+
+						endpointStrings = []
+
+						for experiment in experiments
```
brianbolt, line 674: We actually have the Experiment class and State classes I mentioned on the server side as well.

I think this would be simplified by using Backbone methods and our class Methods like I mentioned above:

https://github.com/mcneilco/acas/blob/7eba2bfe3ff6ec0aa039ab775cc0b045857b6ef1/modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee#L1646

brianbolt (changes requested):
Overall it looks great. I'd just like to see some native Class usage and Backbone usage to avoid duplicating logic around ignored states and ignored values.

bffrost (approved):
This LGTM as-is. Not sure if you're still working on other outstanding items, but if not I think this is good to merge.

You've made lots of big improvements in the structure of the code to reduce duplication and create meaningful, understandable helper functions. Nice work.

modules/ServerAPI/src/client/Protocol.coffee
hhan-schrodinger: Done - thanks for the explanation! Knowing this will be really helpful for me going forward.

modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
hhan-schrodinger: @brianbolt I tried to implement the methods and classes on the server side too but wasn't able to get the getStatesByTypeAndKind() function to work (got ` 'Caught exception: TypeError: experiment.get(...).getStatesByTypeAndKind is not a function\n"`)...do you know how I might be able to get the function in there?

brianbolt: I'd have to dig into the code a bit more than I have time for right now.  TLDR; Lets just mark this as something to possibly come back to. I don't think it's something we need to have if the code is working.

Notes:
Assuming your code snippet says something like `experiment.get('lsStates'). getStatesByTypeAndKind`. If `getStatesByTypeAndKind` is not a function, then it might not have been set as a [StateList](https://github.com/mcneilco/acas/blob/040800fdb4b7b7ec4405d5930ba61b7e16219801/modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee#L756) class.  I might just be a plain dict object.  I would expect though if `experiment` had been created as an instance of [Experiment](https://github.com/mcneilco/acas/blob/040800fdb4b7b7ec4405d5930ba61b7e16219801/modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee#L1646) then it would have a [StateList](https://github.com/mcneilco/acas/blob/040800fdb4b7b7ec4405d5930ba61b7e16219801/modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee#L1657) property

brianbolt (approved):
LGTM.  Thanks for researching using Backbone models but let's just move ahead as is.

modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
hhan-schrodinger: Yep, I was using it `experiment.get('lsStates').getStatesByTypeAndKind`, with a line before it explicitly calling `experiment = new Experiment experiment ` and it was throwing the error. 

After your comment I also tried to also break it up with `experimentLsStates = new StateList experiment.get('lsStates')` and then calling `experimentLsStates.getStatesByTypeAndKind` which still threw the same error?

PR #1052, opened 2022-12-13
ACAS-565: Fix curve render color by

brianbolt:
## Description
**Steps to Reproduce:**

- Load a dose response file into ACAS
- Grab a curve id using the following query `select curveid from api_curve_params limit 1;`
- Render the curve URL:
http://localhost:3000/api/curve/render/dr?curveIds=AG-00000017_3&colorBy=protocol

**Expected Results:**
The plot renders with the protocol name like this:

AG-00000017_3-1

**Actual Results:**
The plot renders but the curve ID (an random identifier) is shown in the plot

AG-00000017_3

**Additional info:**
Also affected by this bug would be plotting flat curves for inactive or highly active curves when fit using the ACAS dose response fit (would not apply to curves just uploaded without an ACAS curve fit)

## Related Issue
ACAS-565

## How Has This Been Tested?
Verified that legend shows the correct text when set to batch, protocol or experiment and shows the curve id when colorBy is not given in the URL

### DIFF
--- a/modules/CurveAnalysis/src/server/renderCurve.R
+++ b/modules/CurveAnalysis/src/server/renderCurve.R
@@ -28,18 +28,17 @@ renderCurve <- function(getParams, postData) {
   # Get protocol curve min and max
   protocol_display_values <- racas::get_protocol_curve_display_min_and_max_by_curve_id(parsedParams$curveIds[[1]])
 
-  output <- applyParsedParametersToFitData(fitData, parsedParams, protocol_display_values)
-  data <- output$data
-  parsedParams <- output$parsedParams
-
-  fitData[ , renderingOptions := list(list(get_rendering_hint_options(renderingHint))), by = renderingHint]
-
-  data <- list(parameters = as.data.frame(fitData), points = as.data.frame(rbindlist(fitData$points)))
-
   #Retrieve rendering hint parameters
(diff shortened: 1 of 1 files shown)


modules/CurveAnalysis/src/server/renderCurve.R
```diff
@@ -28,18 +28,17 @@ renderCurve <- function(getParams, postData) {
   # Get protocol curve min and max
   protocol_display_values <- racas::get_protocol_curve_display_min_and_max_by_curve_id(parsedParams$curveIds[[1]])
 
-  output <- applyParsedParametersToFitData(fitData, parsedParams, protocol_display_values)
-  data <- output$data
-  parsedParams <- output$parsedParams
-
-  fitData[ , renderingOptions := list(list(get_rendering_hint_options(renderingHint))), by = renderingHint]
-
-  data <- list(parameters = as.data.frame(fitData), points = as.data.frame(rbindlist(fitData$points)))
```
brianbolt, line 37: The main issue was here.  Basically wiping out the data variable which eliminated any of the work applied in `applyParsedParmatersToFitData`

bffrost (approved):
LGTM

jmcneil86043 (approved):
I applied this change by hand to all the Nimbus servers, it works great!

brianbolt:
Great thanks. Merging now.

PR #1053, opened 2022-12-14
Protocol Endpoint Manager: Disable edit & removal of endpoint if endpoint is still used on any experiments

hhan-schrodinger:
## Description
Currently, users can delete or edit endpoints which are still being used by any non-deleted experiments, which could be problematic.

To fix this we:
- Added logic to the Protocol Editor that once the page is loaded, will check which endpoints are in use by experiments and disable all fields and the remove button for that row in the endpoint manager
- Added tooltips that explain why the row is disabled when hovering with the mouse over a disabled input field or button

*This was built while also making changes in ACAS-289 (moving all protocol experiment data to @protocolExperiments instead of making multiple calls), so the target branch has been set to ACAS-289 to improve readability of this PR*

Other:
- Fixed misaligned "Download All Files" button for experiments associated with the protocol endpoint manager

## How Has This Been Tested?
Tested locally using a protocol that contains endpoints that are in use with some experiments and also endpoints that are not in use. 

Disabled rows:
Screen Shot 2022-12-14 at 5 28 56 PM

Tooltip on hover explaining why field is disable:
Screen Shot 2022-12-14 at 6 50 12 PM


modules/ServerAPI/src/client/Protocol.coffee
```diff
@@ -775,18 +802,16 @@ class EndpointListController extends AbstractFormController
...
-				@$(".bv_remove_row").hide()
 
 		if @options.readOnly == false && @options.newProtocol == false 	#Only render experiments if the protocol is not new and is not read only 
 			@getExperimentsForProtocol()
 		else #if the table isn't rendered, don't render the download files button either
 			@$(".bv_downloadFiles").hide() 
+			if @options.view == "experiment"
+				@getEndpointTable()
+
+			# Need to add handling for rendering endpoint table if it is an experiment editor/browser and not a protocol browser.
```
bffrost, line 813: nit: Looks like you did this so you can remove this commented out bit

modules/ServerAPI/src/client/Protocol.coffee
```diff
@@ -1127,10 +1184,59 @@ class EndpointListController extends AbstractFormController
...
+			if lsValue.lsKind == "column time" and lsValue.ignored == false and lsValue.numericValue != undefined
+				endpointTime = lsValue.numericValue
+			if lsValue.lsKind == "column time units" and lsValue.ignored == false and lsValue.codeValue != undefined
+				endpointTimeUnits = lsValue.codeValue
+			if lsValue.lsTypeAndKind == "codeValue_hide column" and lsValue.ignored == false and lsValue.codeValue != undefined
+				endpointHidden = lsValue.codeValue
+			
+			endpointStr = endpointName + endpointUnits + endpointDataType + String(endpointConc) + endpointConcUnits + String(endpointTime) + endpointTimeUnits + endpointHidden
+		
+		endpointData =
```
bffrost, line 1224: nit: I guess this works without an explicit `return` but I personally think it'd be clearer if this function had a 
`return endpointData` at the bottom

bffrost (approved):
This looks good to me!

Left a few nits, and I think there are some other nits that could slightly clean up the fetching of values from experiments, but I think this code is good as-is.

modules/ServerAPI/src/client/Protocol.coffee
hhan-schrodinger: That wasn't intentional! Weird that it works without it...

modules/ServerAPI/src/client/Protocol.coffee
hhan-schrodinger: Good catch, thanks!

PR #1054, opened 2022-12-16
Regain debugger attachment on java 11

bffrost:
## Description
Local dev environment java debugging stopped working with Java 11 upgrade.
Found an easy fix: See https://www.baeldung.com/java-application-remote-debugging#from-java9

## How Has This Been Tested?
- Confirmed I can attach a debugger after making this change


brianbolt (approved):
LGTM

PR #1058, opened 2023-01-04
ACAS-574: Enable endpoint manager by default

bffrost:
## Description
The endpoint manager feature is ready for prime time testing, so it should be on by default.
For current python tests to pass, strict matching default must be false. This is also the "least breaking change" behavior, so it is a sensible default.

## How Has This Been Tested?
Local unit tests, will also be watching the CI tests on this PR


dalejerikson (approved):
👍

PR #1060, opened 2023-01-05
ACAS-575: Add parameter to experiment search to filter by project code

brianbolt:
## Description
Checks for projectCodes query parameter in experiment search route
Passes the intersections of allowed projects and requested projectCodes to persistence layer for filtering experiments by project

## Related Issue
ACAS-575

## How Has This Been Tested?
Ran acasclient tests


bffrost (approved):
LGTM, this is an API-only feature right now, not used by the GUI

PR #1064, opened 2023-01-10
route '/api/cmpdRegBulkLoader/getLotsByBulkLoadFileID/:bulkFileID

brianbolt:


dalejerikson (approved):
👍

modules/CmpdRegBulkLoader/src/server/routes/CmpdRegBulkLoaderRoutes.coffee
```diff
@@ -362,9 +364,49 @@ class HTTPResponseError extends Error
...
+		resp.set({
+			"content-length": response.headers.get('content-length'),
+			"content-type": response.headers.get('content-type'),
+		})
+		response.body.pipe(resp);
+	catch err
+		console.log "Error calling service: " + err
+		resp.statusCode = 500
+		resp.json err
+
```
bffrost, line 409: This makes sense as a refactor to use your new `exports.getLotsByBulkLoadFileIDInternal`, but why is the old implementation still present below? Wouldn't we want to remove that?

bffrost (changes requested):

modules/CmpdRegBulkLoader/src/server/routes/CmpdRegBulkLoaderRoutes.coffee
brianbolt: Thanks, missed that checkin, just pushed.

bffrost (approved):
LGTM

PR #1065, opened 2023-01-10
ACAS-580: Fix for blocking registration in UI

brianbolt:
## Description
 - Fix for blocking registration when system needs restandardization

## Related Issue
ACAS-580

## How Has This Been Tested?
Deployed updated code to a server which should have been blocking registration and verified this fixes the issue.  Ran standardization and verified the user message message and blocking went away.


dalejerikson (approved):
👍

PR #1068, opened 2023-01-18
Endpoints are editable under Experiment Editor when Experiment is in "In Process" status

hhan-schrodinger:
## Description
When the experiment status in the experiment editor was changed from "Approved" or "Rejected" to "Created"/"In Process"/"Complete", the fields in the endpoint table were editable when they should not be editable in the experiment editor. 

This was being caused by logic that would enable or disable all fields in the page when the status was changed from a locked status ("Approved"/"Rejected") to an unlocked status ("Created"/"In Process"/"Complete"). 

To fix this, I added some lines to the function that handles when the status is changed that will always disable the endpoint table fields (I brought a copy of the function from BaseEntityController so that I wouldn't change anything universally). 

## How Has This Been Tested?
Tested it by toggling between the different statuses of an experiment. 
[](https://user-images.githubusercontent.com/107148842/213209873-6f9f1957-6b85-4ed7-a15f-d5233e4c0898.mov)


bffrost (approved):
LGTM! Nice clean fix

PR #1070, opened 2023-01-25
Bug Fix: Adding New Lot Copies Lot Corp Name and Causes Error in Subsequent Search Query 

dalejerikson:
## Description

When adding a new lot, the  compound ID that gets copied over includes the lot suffix which results in the 'Next Step' failing (see screenshot).

Screen Shot 2023-01-25 at 10 56 34 AM

If the lot suffix is dropped then the subsequents steps function as normal. 

This fix includes code to drop the lot suffix and prevents it from being copied over whenever a "New Lot" button is utilized. 

Screen Shot 2023-01-25 at 10 57 57 AM

The second step now is successful. 

Screen Shot 2023-01-25 at 10 59 11 AM

NOTE: This previously worked on an environment where ACASLabelSequences is not utilized. 

## How Has This Been Tested?
Manually

### DIFF
--- a/modules/CmpdReg/src/client/src/MetaLot.js
+++ b/modules/CmpdReg/src/client/src/MetaLot.js
@@ -433,7 +433,7 @@ $(function() {
 
 	    newLot: function () {
 			corpName = this.model.get('parent').get('corpName');
-		    window.open("#register/" + corpName);
+			window.open("#register/" + corpName);
 	    },
 
 	    downloadLot: function() {
--- a/modules/CmpdReg/src/client/src/MetaLot.js
+++ b/modules/CmpdReg/src/client/src/MetaLot.js
@@ -431,7 +431,12 @@ $(function() {
 	    },
 
 	    newLot: function () {
-		    window.open("#register/" + this.lotController.model.get('corpName'));
+			// Need to strip away lot suffix to prevent error on Cmpd ID search 
+			corpName = this.lotController.model.get('corpName'); 
+			corpNameArr = corpName.split("-");
+			corpNameArr.pop(); // Removes lot suffix from array 
+			corpName = corpNameArr.join("-");
+		    window.open("#register/" + corpName);
 	    },
 
 	    downloadLot: function() {
--- a/modules/CmpdReg/src/client/src/MetaLot.js
+++ b/modules/CmpdReg/src/client/src/MetaLot.js
@@ -365,6 +365,7 @@ $(function() {
 				var newLotSuccessController = new NewLotSuccessController({
 					el: this.$('.NewLotSuccessView'),
 					corpName: message.metalot.lot.corpName,
+					parentCorpName: message.metalot.lot.parent.corpName, 
 					buid: message.metalot.lot.buid
 				});
 				newLotSuccessController.render();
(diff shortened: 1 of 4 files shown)


dalejerikson:
Note: Spacing was messed up in VSCode. Will fix shortly.

dalejerikson:
I was also wondering if making this code change in 3 places was fine considering the size/simplicity or if I should DRY. Additionally, for the purposes of testing, this change worked well. However, I'm not 100% sure if the location of this change is appropriate or if there's a better place to do this text processing.

modules/CmpdReg/src/client/src/MetaLot.js
```diff
@@ -431,7 +431,12 @@ $(function() {
 	    },
 
 	    newLot: function () {
-		    window.open("#register/" + this.lotController.model.get('corpName'));
```
bffrost, line 434: It's a little risky to split on "-" since the separators between prefix, parent number, and lot number are all configurable. Can you get the `parent` from somewhere within the `lotController.model` so you can directly pull the parent corpName rather than doing string manipulation on the lot corpName?

bffrost (changes requested):

modules/CmpdReg/src/client/src/MetaLot.js
dalejerikson: Tried to add your feedback to each of the affected areas. Had to do a little differently in each place.

modules/CmpdReg/src/client/src/SearchResults.js
```diff
@@ -74,7 +74,7 @@ $(function() {
 
         newLot: function() {
 	        if ( !window.configuration.metaLot.saltBeforeLot && this.model.get('lotIDs').length==1) {
-		        this.trigger('newLot', this.model.get('lotIDs')[0].corpName);
+		        this.trigger('newLot', this.model.get('corpName'));
```
dalejerikson, line 77: Is doing this fine? I'm not sure if there would be repercussions elsewhere.

modules/CmpdReg/src/client/src/MetaLot.js
```diff
@@ -431,7 +432,8 @@ $(function() {
 	    },
 
 	    newLot: function () {
```
dalejerikson, line 434: I'm a little confused on when this newLot function is used. I understand where the ones in `NewLotSuccess.js` and `Search.js` are used?

bffrost: I think this is tied to the "New Lot" button at the bottom of the Lot detail view of an existing lot:
Screen Shot 2023-01-26 at 4 05 19 PM

modules/CmpdReg/src/client/src/SearchResults.js
bffrost: As long as you've tested the New Lot button from within Search Results locally, and tested it with ACASLabelSequence and "standard" modes, I think you're all set to make this change.

bffrost (approved):
New round of changes LGTM!

modules/CmpdReg/src/client/src/SearchResults.js
dalejerikson: I tested with the search results locally with both ACASLabelSequence and "standard" modes so sounds good

PR #1072, opened 2023-01-26
Mismatch between load experiment instructions and browse select behavior

hhan-schrodinger:
## Description
Instructional text for 'Load Experiment' says to 'select a file', but the file-browser GUI is multi select. This has caused issues where users may think they are uploading many experiments but are only uploading one. 

To fix this, the "Browse Files..." button only allows users to select one file while uploading. Since users can also drag and drop multiple instead of using the "Browse Files..." button, logic was put in place to block validation when multiple files are trying to be uploaded. A error message explains that only one file can be uploaded at a time when this is attempted. 

## How Has This Been Tested?
Tried to use the "Browse Files..." button to select multiple files (no longer possible). Dragged and dropped multiple files and tried to upload them and saw that validation/upload was blocked. Subsequently tried with single files to ensure that the uploading and validation still worked.

### DIFF
--- a/modules/Components/src/client/LSFileChooserView.html
+++ b/modules/Components/src/client/LSFileChooserView.html
@@ -6,9 +6,6 @@
 					<span>Browse Files...</span>
 					<input type="file" name="files[]" multiple>
 				</span>
-                <br/>
-                <br/>
-                <div class="bv_multipleFileWarning hide" style="color: red; font-size: 12px; text-align: left;">Sorry, you can only upload one file at a time. Please choose one file and try again.</div>
 			</div>
 			<div class="span7 well <%- dropZoneClassId %>" style="display: none; text-align: center">
 				<h4 class='bv_dragOverMessage'><%- dragOverMessage %></h4>
--- a/modules/Components/src/client/LSFileChooserView.html
+++ b/modules/Components/src/client/LSFileChooserView.html
@@ -4,7 +4,7 @@
 			<div class="span7 bv_manualFileSelect" style="padding-left:50px;">
 				<span class="btn fileinput-button">
 					<span>Browse Files...</span>
-					<input type="file" name="files[]">
+					<input type="file" name="files[]" multiple>
 				</span>
                 <br/>
                 <br/>
--- a/modules/Components/src/client/LSFileChooserView.html
+++ b/modules/Components/src/client/LSFileChooserView.html
@@ -6,6 +6,9 @@
 					<span>Browse Files...</span>
 					<input type="file" name="files[]">
 				</span>
+                <br/>
+                <br/>
+                <div class="bv_multipleFileWarning hide" style="color: red; font-size: 12px; text-align: left;">Sorry, you can only upload one file at a time. Please choose one file and try again.</div>
 			</div>
 			<div class="span7 well <%- dropZoneClassId %>" style="display: none; text-align: center">
 				<h4 class='bv_dragOverMessage'><%- dragOverMessage %></h4>
--- a/modules/Components/src/client/LSFileChooserView.html
+++ b/modules/Components/src/client/LSFileChooserView.html
@@ -4,7 +4,7 @@
 			<div class="span7 bv_manualFileSelect" style="padding-left:50px;">
 				<span class="btn fileinput-button">
 					<span>Browse Files...</span>
-					<input type="file" name="files[]" multiple>
+					<input type="file" name="files[]">
 				</span>
 			</div>
 			<div class="span7 well <%- dropZoneClassId %>" style="display: none; text-align: center">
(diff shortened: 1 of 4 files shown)


modules/GenericDataParser/src/client/GenericDataParser.coffee
```diff
@@ -8,3 +8,30 @@ class GenericDataParserController extends BasicFileValidateAndSaveController
 		@loadImagesFile = true
 		super()
 		@$('.bv_moduleTitle').html("Simple #{window.conf.experiment.label} Loader")
+
+	validateParseFile: =>
+		
+		# check the number of files uploaded by checking the number of rows in the table of uploaded files
+		# if it is more than one, we want to block the validation 
+		filesUploaded = @$(".files tr.template-download").length
```
brianbolt, line 16: Seems like all of this logic should be handled down the LSFileChooserController where it [detects that it has reached the max number of files] https://github.com/mcneilco/acas/blob/a78c855f3793e981fb58ad4cfb17c9c954ae788f/modules/Components/src/client/LSFileChooser.coffee#L104

brianbolt (changes requested):

brianbolt (commented):
Also want to note, that the (really old) file upload widget we are using actually has an option for maxNumberOfFiles.  I don't now why we aren't using it, but it might be worth a shot [passing the argument to it here](https://github.com/mcneilco/acas/blob/a78c855f3793e981fb58ad4cfb17c9c954ae788f/modules/Components/src/client/LSFileChooser.coffee#L144) and see what happens

https://github.com/blueimp/jQuery-File-Upload/wiki/Options#maxnumberoffiles

modules/GenericDataParser/src/client/GenericDataParser.coffee
hhan-schrodinger: @brianbolt I thought so too and actually started with that (and also the maxNumberOfFiles option) in the LSFileChooserController but didn't have any luck with toggling it. I observed some odd behavior that I couldn't figure out so opted for this current approach. Basically the currentNumberOfFiles variable which is compared against maxNumberOfFiles would go to zero every time, so setting the maxNumberOfFiles wouldn't matter.

modules/Components/src/client/LSFileChooserView.html
```diff
@@ -4,8 +4,11 @@
 			<div class="span7 bv_manualFileSelect" style="padding-left:50px;">
 				<span class="btn fileinput-button">
 					<span>Browse Files...</span>
-					<input type="file" name="files[]" multiple>
+					<input type="file" name="files[]">
```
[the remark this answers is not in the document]
hhan-schrodinger: Moved the logic to GenericDataParser.coffee so it will remove the multiple attribute when the experiment loader is opened, that way it won't affect other parts interfaces.

hhan-schrodinger:
> Also want to note, that the (really old) file upload widget we are using actually has an option for maxNumberOfFiles. I don't now why we aren't using it, but it might be worth a shot [passing the argument to it here](https://github.com/mcneilco/acas/blob/a78c855f3793e981fb58ad4cfb17c9c954ae788f/modules/Components/src/client/LSFileChooser.coffee#L144) and see what happens
> 
> https://github.com/blueimp/jQuery-File-Upload/wiki/Options#maxnumberoffiles

Gave it a shot, no luck sadly!

modules/GenericDataParser/src/client/GenericDataParser.coffee
```diff
@@ -8,3 +8,34 @@ class GenericDataParserController extends BasicFileValidateAndSaveController
 		@loadImagesFile = true
 		super()
 		@$('.bv_moduleTitle').html("Simple #{window.conf.experiment.label} Loader")
+
+		# set the number of uploads possible to one
+		@$(".fileinput-button input").attr("multiple", false)
```
brianbolt, line 13: I believe this is just going to set an attribute on the input and set its value to "false".  I don't think it will remove the attribute which I think is required to make it so you can't select multiple.

brianbolt: In any case, can we move this down into the LSFileChooser controller so that if someone sets max number of files to 1 then it removes the multiple attribute?

brianbolt:
Setting `maxNumberOfFiles: 1` on the file upload widget worked for me.

Screenshot 2023-01-26 at 1 40 30 PM
Screenshot 2023-01-26 at 1 40 42 PM

brianbolt (changes requested):

hhan-schrodinger:
> Setting `maxNumberOfFiles: 1` on the file upload widget worked for me.
> 
> Screenshot 2023-01-26 at 1 40 30 PM
> 
> Screenshot 2023-01-26 at 1 40 42 PM

Oof, realized I had a typo and wrote mmaxNumberOfFiles which is why it didn't work. Thanks!

brianbolt (approved):
LGTM

brianbolt:
> @hhan-schrodinger are you waiting on a review from @brianbolt ?
Not anymore.

PR #1073, opened 2023-01-26
Improving Reporting of Overwritten Experiments 

dalejerikson:
Users often are hampered by apparent 'missing data' when a registrar fails to update the experiment name field.  A new experiment code is generated and the previous ACAS file-data is replaced (overwritten).  In most cases, the current behavior is desired, since it allows trivial adjustments to units, endpoint labels, etc.  But in wrong hands it leads to a confusing state.  Moreover, as registrars it is difficult to even discover such cases and troubleshoot.  Here we want to enhance the Experiment Browser table such that it can be used to recover from such registration failures. 

Scope of This Work 

- Adding "Overwritten" as an "Experiment Status"
- Show Overwritten Experiments in Experiment Browser if the user has client.entity.viewDeletedRoles
- Adding a route in the delete experiment logic that changes the status field 
- Make sure overwritten experiments aren't editable or deletable or openable

### DIFF
--- a/modules/ServerAPI/src/server/routes/ExperimentServiceRoutes.coffee
+++ b/modules/ServerAPI/src/server/routes/ExperimentServiceRoutes.coffee
@@ -529,7 +529,6 @@ exports.genericExperimentSearch = (req, res) ->
 				# Filter the allowed projects by the requested projects
 				allowedProjectCodes = _.intersection(allowedProjectCodes, requestedProjectCodes)
 
-			#baseurl = config.all.client.service.persistence.fullpath+"experiments/search?q="+req.params.searchTerm+"&projects=#{encodeURIComponent(allowedProjectCodes.join(','))}"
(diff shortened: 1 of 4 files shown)


modules/ServerAPI/src/server/routes/ExperimentServiceRoutes.coffee
```diff
@@ -516,7 +529,8 @@ exports.genericExperimentSearch = (req, res) ->
 				# Filter the allowed projects by the requested projects
 				allowedProjectCodes = _.intersection(allowedProjectCodes, requestedProjectCodes)
 
-			baseurl = config.all.client.service.persistence.fullpath+"experiments/search?q="+req.params.searchTerm+"&projects=#{encodeURIComponent(allowedProjectCodes.join(','))}"
+			#baseurl = config.all.client.service.persistence.fullpath+"experiments/search?q="+req.params.searchTerm+"&projects=#{encodeURIComponent(allowedProjectCodes.join(','))}"
```
dalejerikson, line 532: Will remove this

bffrost (approved):
LGTM

PR #1074, opened 2023-02-03
Units in experiment endpoint table are not loading

hhan-schrodinger:
## Description
When an experiment with new column units was uploaded and had an existing protocol, the units were not present in the endpoint table when the experiment was viewed in the experiment browser/editor. 

This was being caused because codetables for new column data were only saved to the database in the case that the experiment was associated with a new protocol. To fix, logic was moved so that codetables would be saved on any valid experiment load. 

## How Has This Been Tested?
Tested by loading an experiment with new column units against an existing protocol. Units were displayed in the experiment when viewed in the experiment browser/editor.


bffrost (approved):
LGTM!

PR #1075, opened 2023-02-03
Experiment search in endpoint manager not filtering properly

hhan-schrodinger:
## Description
In the protocol endpoint manager, users should be able to click on a row in the endpoint table find experiments using that protocol that have columns with the corresponding data. 

It was observed that when using this search feature some experiments that should have been showing up in the results were not, and conversely, experiments that should not be showing up were. This was being caused by logic that was not clearing the variables used to see if values matched between searches. The error was fixed by resetting variables each round of searching. 

## How Has This Been Tested?
Tested locally with a set of experiments with different units, column names loaded on one protocol to see if searches would load the correct set of experiments.


bffrost (approved):
LGTM

PR #1076, opened 2023-02-10
Endpoint rows are still disabled after experiments are deleted

hhan-schrodinger:
## Description
In the protocol endpoint manager, endpoint rows are disabled whenever the endpoint is in use by an experiment to prevent discrepancies from occurring. However, it was noticed that the endpoint rows were still disabled when all experiments associated with the protocol using the endpoint were deleted.

This was because there were no steps to ensure that deleted and ignored experiment endpoints were not being considered by the logic that was used to disable the rows. To fix this, a step was added to filter out deleted and ignored experiments. 

## How Has This Been Tested?
Replicated the original issue locally by uploading a protocol, deleting the experiments and seeing if the rows in the protocol endpoint manager were now enabled (which they were).


bffrost (approved):
LGTM, short and sweet.

dalejerikson (approved):
nice clean code :)

PR #1077, opened 2023-02-10
Download Template button in protocol browser/editor not working

hhan-schrodinger:
## Description
It was observed that the "Download Template" button in the protocol browser and protocol editor was not working. 

This was caused by logic when creating the .csv template file that could not handle instances where the protocol did not have a project assigned. To fix, logic was added to handle cases where there was no project for a protocol. 

## How Has This Been Tested?
Tested locally with protocols where the "Download Template" button was not previously working and now worked with the added logic.


dalejerikson (approved):
👍

bffrost (approved):
Makes sense! LGTM

PR #1078, opened 2023-02-10
Update code to experimentCode & name to experimentName

iamanandkr:
## Description

In [ACAS-307 | acas-roo-server](https://github.com/mcneilco/acas-roo-server/pull/422) the `/api/v1/metalots/checkDependencies/corpName/{corpName}` route has changed the keys in the `linkedExperiments`
- code > experimentCode
- name > experimentName

`getLotDependenciesInternal` is the function which fetches the lot dependencies and does post processing on the data. Here the code updates the key information across usages.

## Related Issue

[ACAS-307](https://jira.schrodinger.com/browse/ACAS-307)
## How Has This Been Tested?

- I ran all the acasclient tests
- I verified via the acasclient the API new updated values:
```json
    "batchCodes": [
        "CMPD-0000001-001"
    ],
    "linkedContainers": [],
    "linkedDataExists": true,
    "linkedExperiments": [
        {
            "comments": "CMPD-0000001-001",
            "description": "60 results and 280 raw results",
            "experimentCode": "EXPT-00000001",
            "experimentName": "4 parameter D-R - 2018-05-08",
            "protocolCode": "PROT-00000001",
            "protocolName": "Dose Response Protocol",
            "acls": {
                "read": true,
                "write": true,
                "delete": true
            }
        }
    ],
```

PR #1079, opened 2023-02-14
Add protocol and analysis group codes to experiments linked to a lot.

iamanandkr:
## Description

I have [acas-roo-server: PR](https://github.com/mcneilco/acas-roo-server/pull/422) and [acas: PR](https://github.com/mcneilco/acas/pull/1078) to add more experiment information to the linked experiments of a Lot but I think I could have done it more simply by querying and adding the fields in the acas layer instead of writing specific SQL queries and creating DTO for the data format.

Note:
If we would want the same information to be available across all usages of the `CmpdRegBatchCodeDTO` then it would be reasonable to add all the information in the acas-roo layer but I thought of trying this and getting feedback.

- I have refactored the code to create a separate function that handles ACLs and adds protocol and analysis group information to the experiment object.
- I fixed a bug with `fetchExperimentsByCodeNames` where the URL wasn't getting formatted correctly because of missing "?" between the host and the query params.

## Related Issue

[ACAS-307](https://jira.schrodinger.com/browse/ACAS-307)
## How Has This Been Tested?

```json
"linkedExperiments": [
    {
        "code": "EXPT-00000001",
        "comments": "CMPD-0000001-001",
        "description": "60 results and 280 raw results",
        "ignored": false,
        "name": "4 parameter D-R - 2018-05-08",
        "acls": {
            "read": true,
            "write": true,
            "delete": true
        },
        "analysisGroups": [
            {
                "codeName": "AG-00000006",
                "values": [
                    {
                        "id": 190,
                        "lsKind": "Key",
                        "lsType": "numericValue",
                        "value": 6
                    },
                    {
                        "id": 193,
                        "lsKind": "Rendering Hint",
                        "lsType": "stringValue",
                        "value": "4 parameter D-R"
                    },
                    {
                        "id": 191,
                        "lsKind": "Max",
                        "lsType": "numericValue",
                        "value": 98.6500950200081
                    },
                    {
                        "id": 194,
                        "lsKind": "Slope",
                        "lsType": "numericValue",
                        "value": -0.821450639331544
                    },
                    {
                        "id": 192,
                        "lsKind": "Min",
                        "lsType": "numericValue",
                        "value": 1.50788601214144
                    },
                    {
                        "id": 189,
                        "lsKind": "EC50",
                        "lsType": "numericValue",
                        "value": 0.562143112495009
                    },
                    {
                        "id": 195,
                        "lsKind": "curve id",
                        "lsType": "stringValue",
                        "value": "f_AG-00000006"
                    }
                ]
            }
        ],
        "protocol": {
            "code": "PROT-00000001",
            "name": "Dose Response Protocol"
        }
    }
]
```

## TODO:
~Add acasclient tests.~

### DIFF
--- a/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
+++ b/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
@@ -282,17 +282,14 @@ exports.getLotDependenciesByCorpNameInternal = (lotCorpName, user, allowedProjec
 	dependencies = await exports.getLotDependenciesInternal(metaLot.lot, user, allowedProjects, includeLinkedLots)
 	return dependencies
 
-exports.getLotDependenciesInternal = (lot, user, allowedProjects, includeLinkedLots=true) ->
-	console.log "Checking lot dependencies for lot #{lot.corpName} with user #{user.username}"
-
-	if !allowedProjects?
-		[err, allowedProjects] = await serverUtilityFunctions.promisifyRequestStatusResponse(authorRoutes.allowedProjectsInternal, [user])
-		if err?
-			throw new InternalServerError "Error checking user projects"
-			return
-
+###*
+# Get the dependencies for the lot. Dependencies will account for user ACLs.
+# @param {string} lotCorpName - Lot corporate name.
+# @param {object} user - User object.
+# @param {array} allowedProjects - Array of projects the user has access to.
+###
+getLotExperimentDependencies = (lot, user, allowedProjects) ->
 	lotCorpName = lot.corpName
-
 	# Get the depdencies from the service which does not cover user ACLS
 	response = await exports.fetchMetaLotDependencies(lotCorpName)
 
@@ -313,7 +310,7 @@ exports.getLotDependenciesInternal = (lot, user, allowedProjects, includeLinkedL
 		experimentCodeList = _.uniq experimentCodeList
 
 		# Get the experiments from the server
-		response = await experimentServiceRoutes.fetchExperimentsByCodeNames(experimentCodeList)
+		response = await experimentServiceRoutes.fetchExperimentsByCodeNames(experimentCodeList, "analysisgroups")
 		experiments = await response.json()
 
 		# It's unexpected that the server would return a list of experiments that are not in the list of codes we asked for, so we check for that and erorr
@@ -325,7 +322,7 @@ exports.getLotDependenciesInternal = (lot, user, allowedProjects, includeLinkedL
 
 		# Get the acls for the experiments
 		for experiment in experiments
-			console.log "Checking acls for experiment #{experiment.code}"
+			console.log "Checking acls for experiment #{experiment.experiment.codeName}"
 
 			# This returns the acls (read, write, delete of the experiment for the user and allowed project the user has)
 			acls = await experimentServiceRoutes.getExperimentACL(experiment.experiment, user, allowedProjects)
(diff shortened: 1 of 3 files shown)


bffrost (approved):
This looks great @iamanandkr , and I agree it's a more elegant fix.
I prefer this approach to the Roo + ACAS proposal. But curious what @brianbolt says too.

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
```diff
@@ -335,14 +332,38 @@ exports.getLotDependenciesInternal = (lot, user, allowedProjects, includeLinkedL
...
+				# The experiment is readable so include the experiment and acls
+				linkedExperiment = dependencies.linkedExperiments[idx]
+				linkedExperiment.acls = acls
+		
+		# Add protocol and analysis group codes to user accessible experiments
+		for linkedExperiment in dependencies.linkedExperiments
+			if not linkedExperiment.acls.getRead()
+				continue
+			experiment = _.findWhere(experiments, {experimentCodeName: linkedExperiment.code})
+			linkedExperiment.protocolCode = experiment.experiment.protocol.codeName
```
brianbolt, line 348: I think the users may also want to see the protocol name which is buried in the protocol.lsLabels.  Can we pul that up? Perhaps the output should be:

```
"linkedExperiments": [
        {
            "code": "EXPT-00000001",
            "comments": "CMPD-0000001-001",
            "description": "60 results and 280 raw results",
            "ignored": false,
            "name": "4 parameter D-R - 2018-05-08",
            "acls": {
                "read": true,
                "write": true,
                "delete": true
            },
            "protocol": {
                  "code": "PROT-00000001",
                  "name": "BLAHBLAH Protocol"
 ```

File: modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee [hunk 2]
```diff
@@ -335,14 +332,38 @@ exports.getLotDependenciesInternal = (lot, user, allowedProjects, includeLinkedL
...
+			analysisGroupCodes = _.pluck(experiment.experiment.analysisGroups, "codeName")
+			linkedExperiment.analysisGroupCodes = analysisGroupCodes.sort()
```
brianbolt, line 350: I think the [original ticket](https://jira.schrodinger.com/browse/ACAS-307) referenced digging down into the analysis groups -> states -> values and plucking the values like:

```
result type
values 
```

I would think the users would benefit from fetching the analysis group values from unignored analysis group, states and values e.g output:
```
analysis_group_results: [
{ 
    "id": 1
    "lsType": "stringValue"
    "lsKind": "My String Answer"
    "stringValue": "hello"
}, {
{ 
    "id": 2
    "lsType": "numericValue"
    "lsKind": "My Numeric Answer"
    "numericValue": 1234.1234
},
```

brianbolt: @bffrost Does that seem right to you?

We'll probably additionally want to filter out all analysis group values where lsKind = "batch code" as that is where the lot corp name is which is already implied by the output of this data.

brianbolt (changes requested):
I do like the structure of this but I think the output doesn't match the spec on the ticket.  I'm requesting changes to see if you can make this approach match the spec.

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
bffrost: Sorry for the delayed reply, but yes what @brianbolt says is right. @iamanandkr it'd be good to include the analysis group values in the way that Brian showed

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
iamanandkr: Done.

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
iamanandkr: Done.

bffrost (approved):
@iamanandkr this LGTM, but it'd be good for you to post an updated example of the JSON this route responds with, and also get @brianbolt 's approval.

brianbolt (approved):
LGTM. Thanks!

PR #1083, opened 2023-03-14
ACAS-589: Sort PickListList collection items.

iamanandkr:
### Issue
Lot Chemist, Protocol Scientist, Experiment Scientist drop down menu don't show names in alphabetical order. We have here two pick list components.
#### Compound Registration
`Stereo Category`, `Chemist`, `Physical State`, `Purity measured by`, `Vendor`, `Retain` units, `Project`, `Amount` units, `Purity` & `Solution Amount` units  uses [PickListSelectController](https://github.com/mcneilco/acas/blob/a2f657ae8d9d6fccceef561fcbb019fb98d72965/modules/CmpdReg/src/client/src/PickList.js#L48)
`Salt`, `Isotope` uses [IsoSaltEquivListController](https://github.com/mcneilco/acas/blob/a2f657ae8d9d6fccceef561fcbb019fb98d72965/modules/CmpdReg/src/client/src/IsoSaltEquiv.js#L178) which has it's own comparator for the collection.
##### Protocol Editor
`Status`, `Scientist`, `Project`, `Assay Stage`, `Attach Files > File Type`
##### Experiment Editor
`Status`, `Scientist`, `Protocol`, `Project`, `Attach Files > File Type`
##### Example Thing
`Scientist`, `Attach multiple files`
##### Project Editor
`Status`, `Project Leader`, `Attach Files > File Type`
##### Author Editor
`System Roles`, `Project Roles`
##### Label Sequence
`Linked Roles`, `Label Type_Kind`, `Thing Type_Kind`

Data loader options use the [pick list options](https://github.com/mcneilco/acas/blob/a2f657ae8d9d6fccceef561fcbb019fb98d72965/modules/Components/src/client/PickList.coffee#L4). Data loader has can sort items on display orders and on name.

Here I made global changes based on discussion with @bffrost. All the option that we have in the GUI generally make sense to be ordered alphabetically. We do have DDict values which contains `displayOrder` attribute for displaying data and so here the code first checks if the `displayOrder` is an available attribute, else sort on the `name`. 

## Related Issue

[ACAS-589](https://jira.schrodinger.com/browse/ACAS-589)
## How Has This Been Tested?

image

image

image

image

image

### DIFF
--- a/modules/CmpdReg/src/client/src/PickList.js
+++ b/modules/CmpdReg/src/client/src/PickList.js
@@ -9,6 +9,14 @@ $(function() {
 	window.PickListList = Backbone.Collection.extend({
 
 		model: PickList,
+
+		comparator: function(model) {
+			var name = model.get('name');
+			if (name) {
+				return name.toLowerCase();
+			}
+			return name;
+		},
 		
 		url: function() {
 			if(window.configuration.serverConnection.connectToServer) {
--- a/modules/Components/src/client/PickList.coffee
+++ b/modules/Components/src/client/PickList.coffee
@@ -3,6 +3,13 @@ class PickList extends Backbone.Model
 class PickListList extends Backbone.Collection
 	model: PickList
 
+	comparator: (model) ->
+		displayOrder = model.get('displayOrder')
+		if displayOrder?
+			displayOrder
+		else
+			model.get("name")?.toLowerCase()
+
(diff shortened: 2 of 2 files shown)


bffrost (approved):
LGTM, but would be good to get your review @brianbolt

brianbolt (approved):
LGTM

PR #1084, opened 2023-03-16
Add special character handling to SEL template file download

hhan-schrodinger:
## Description
Users of the protocol endpoint manager asked how well the template download feature could handle special characters. To test it, a string of special characters (@"|!#$%&/()=?»«@£§€{}.-;'<>_,") was inserted in different parts of a protocol that was uploaded. When trying to download the template of this protocol, the file was truncated in the middle of the string: 

image

In order to fix, the template file .csv construction was refactored to use the csv-stringify library instead of manually constructing the .csv string. The frontend was also updated to handle the new format. 

A minor bug where the warning/success text for "Download All Files" would also show when hitting the "Download Template" button from the text divs sharing the same class name was fixed as well. 

## How Has This Been Tested?
Tested with the same protocol that was breaking with the original code. While it would no longer break from almost all special characters being included, it can not handle commas (would cause a new cell to appear). 

Screenshot 2023-03-16 at 1 06 26 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,8 @@
     "url-loader": "4.1.1",
     "winston": "3.3.3",
     "winston-mongodb": "5.0.5",
-    "yamljs": "0.3.0"
+    "yamljs": "0.3.0",
+    "csv-stringify": "6.3.0"
   },
   "devDependencies": {
     "coffeescript": "2.5.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,7 +76,8 @@
     "url-loader": "4.1.1",
     "winston": "3.3.3",
     "winston-mongodb": "5.0.5",
-    "yamljs": "0.3.0"
+    "yamljs": "0.3.0",
+    "csv-stringify": "6.3.0"
```
brianbolt, line 80: Generally don't like to add new packages to ACAS when it can be avoided but in this case, I can't find any examples of usage of the other package we have called 'json2csv'. It also seems to be abandoned so it's worth using this one either way.

brianbolt (approved):
LGTM

PR #1085, opened 2023-03-20
ACAS-612: Provide delete_type and delete_url and noop delete file route

brianbolt:
## Description
 - Fixes: The "Browse Files..." button for the query-file-upload wasn't being re-enabled after we started setting maxNumberOfFiles in https://github.com/mcneilco/acas/pull/1072
 -   The jQuery file plugin (which we have customized) requires a `delete_type` and `delete_url` returned from the server when uploading a file or it will not reset the maxNumberOfFiles when you click delete.
 - To fix this I added:
   - The `delete_url` and `delete_type` to our upload route
   - A no-op delete route to data files (Didn't implement actual delete because there are lots of implications we have not fully though through yet.

## Related Issue
ACAS-612

## How Has This Been Tested?
Ran through the UI process for uploading an experiment file (refreshed between each bullet point)
   - Upload file, validate, save, load another, (had to click delete file but I think that's normal), upload new file, validate, save
   - Upload file, validate, back, delete file, upload new file, validate, save
   - Verified that I could not multi select
   - Verified that on each delete, I saw an Ajax request to delete the file which noop (logged in the expected message on the server)

### DIFF
--- a/modules/ServerAPI/src/server/routes/FileServices.coffee
+++ b/modules/ServerAPI/src/server/routes/FileServices.coffee
@@ -92,10 +92,10 @@ setupRoutes = (app, loginRoutes, requireLogin) ->
 							"originalName": file.originalname,
 							"size": file.size,
 							"type": file.mimetype,
-							# Note, the delete_type and delete_url are required by the jquery file upload plugin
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM, thanks for finding this. Not at all intuitive!

PR #1094, opened 2023-03-29
Unexpected time values showing up in downloaded SEL template files

hhan-schrodinger:
## Description
When downloading an SEL file for a protocol with multiple experiments, I observed that random time values were being inserted into the template that were not in any of the experiments or in the protocol endpoints. Even more strangely, the values would be associated with different endpoints and have different values each download. 

The error was coming from a line in the template file generating logic for endpoint time, where a single `=` was used instead of a `==`, so all arguments were being treated as true and random values that should not have passed the `if` criteria were being accepted and inserted. 

## How Has This Been Tested?
Tested downloading template files for the same protocol before and after the change several times and observed no more random values were being inserted in the time section.


brianbolt (approved):
I updated the base branch as we are targeting 2023.2.x.

LGTM.

PR #1100, opened 2023-04-20
Endpoint manager breaks creating new protocols

hhan-schrodinger:
## Description
When making a new protocol with the Protocol Editor, users were unable to add new endpoints and save the protocol. When users clicked on the "Add Endpoint" button, nothing would happen. This was due to an issue where the @endpointControllers object that is part of adding, removing, and editing endpoint rows was not being initialized for new protocols. To fix, added logic specific for new protocols to initialize this. 

## How Has This Been Tested?
After making the change, successfully created and saved new protocols with endpoints using the endpoint editor.


hweissig (approved):
LGTM!

bffrost (approved):
LGTM

brianbolt (approved):
LGTM

PR #1101, opened 2023-04-20
Download template does not fill endpoints when there are no experiments exist on the protocol

hhan-schrodinger:
## Description
When users would try to download template files when a protocol had no experiments associated with it, the download would fail. This is because to generate the template file, the backend logic finds all the endpoints associated with each experiment and compiles them to create the file. It doesn't look at the protocol's saved endpoints, and couldn't handle if there were no endpoints at all.

To fix, implemented new logic:

- Check if the protocol has experiments, if it does - extract the endpoints from the experiments and display them in the file
-  If the protocol doesn't have experiments, use the endpoints defined in the protocol
- If the protocol doesn't have any endpoints, don't add the section to the .csv when creating it

## How Has This Been Tested?
Created three protocols: one with experiments uploaded associated with it, one with no experiments and no endpoints, and one with no experiments but several endpoints. 

Tested downloading template files for each. The first protocol with experiments showed all endpoints in all experiments (expected behavior). The second one with no experiments and no endpoints had experiment metadata but no endpoints in the template file (expected behavior). The third protocol with no experiments but endpoints defined in the protocol showed the endpoints defined in the protocol (expected behavior).

### DIFF
--- a/modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
+++ b/modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
@@ -662,6 +662,21 @@ exports.getTemplateSELFile = (req, resp) ->
 				protocolDate = todayDate.getMonth() + 1 + "/" + todayDate.getDate() + "/" + String(todayDate.getFullYear())[2..4]
 			
 				protocolScientist = req.session.passport.user.username
+
+				# create the top portion of the template file given the protocol metadata just extracted...
+				data = [  ['Experiment Meta Data'],
+				['Format', 'Generic'],
+				['Protocol Name', protocolName],
+				['Experiment Name', ''],
+				['Scientist', protocolScientist],
+				['Notebook', ''],
+				['Page', ''],
+				['Assay Date', protocolDate],
+				['Project', protocolProject],
+				[''],
+				['Calculated Results']
+				]
+
 				experimentServiceRoutes.experimentsByProtocolCodenameInternal protocolCode, false, (statusCode, experiments) ->
 					if statusCode == 200
 						# Part 1: Find all unique endpoints across experiments that use this protocol code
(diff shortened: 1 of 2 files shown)


hweissig (approved):
LGTM!

modules/ServerAPI/src/server/routes/ProtocolServiceRoutes.coffee
```diff
@@ -662,153 +662,174 @@ exports.getTemplateSELFile = (req, resp) ->
...
+							# if the protocol has no endpoints saved in it, we record it has no endpoint data and won't write that section of the .csv...
+							if protocolEndpointLsStates.length == 0
+								protocolHasEndpoints = false 
+
+							# ...but if it does, we want to extract them and format them for the template file
+							else
+								for lsState in protocolEndpointLsStates
+									extractEndpointData(lsState)
+
+						# if the protocol has experiments, collect all the endpoint information available
```
bffrost, line 754: @hhan-schrodinger @brianbolt Do we expect the experiment endpoints to differ from the protocol endpoints? If so, wouldn't we want the template to reflect the curated endpoints defined on the protocol?

I don't see a reason to check the experiments at all - rather I think we shuld just be reading the `protocolEndPointLsStates`. Can either of you think of a reason why we should be looking at the experiments? It's possible I'm missing something here.

brianbolt: If you create a protocol when uploading an experiment via Experiment Loader (SEL) does it create the protocolEndPointLsStates?  If not, then we should add it and then I don't think there is every a reason to look at the experiment level when using the protocol editor.

hhan-schrodinger: I think that is much simpler to have the template just reflect endpoints defined on the protocol too. I vaguely remember looking across all the endpoints in all associated experiments being the original intention maybe to find all possible endpoints for a protocol (maybe to look for erroneous endpoints that wouldn't show up in the endpoint manager)?

bffrost: @hhan-schrodinger Let's simplify it just to look at the protocol endpoints. Less code means less places for bugs to hide.

hhan-schrodinger: @bffrost Sounds good, made the changes!

bffrost (approved):
LGTM! Thanks for the changes @hhan-schrodinger

brianbolt (approved):
LGTM

PR #1102, opened 2023-04-21
1096 read optional experiment status from sel file header block

philippcheung:

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -4730,7 +4730,3 @@ createColumnOrderStates <- function(exptDataColumns=selColumnOrderInfo, errorEnv
     }
     return(experimentStates)
 }
-
-
-
-
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -2170,11 +2170,13 @@ createNewExperiment <- function(metaData, protocol, lsTransaction, pathToGeneric
     codeType = "assay",
     codeKind = "scientist",
     lsTransaction= lsTransaction)
-  # experimentStatus <- racas::applicationSettings$server.sel.experimentStatus
   experimentStatus <- metaData$"Experiment Status"
   if (is.null(experimentStatus) || experimentStatus == "") {
-    experimentStatus <- "approved"
+    experimentStatus <- racas::applicationSettings$server.sel.experimentStatus
   }
+  if (is.null(experimentStatus) || experimentStatus == "") {
+    experimentStatus <- "approved"
+  }  
   experimentValues[[length(experimentValues)+1]] <- createStateValue(
     recordedBy = recordedBy,
     lsType = "codeValue",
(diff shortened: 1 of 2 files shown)


jmcneil86043 (approved):
LGTM

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -2372,6 +2380,61 @@ validateProject <- function(projectName, configList, username, protocolName = NU
...
+  #   testMode:               If true, the function bypasses Roo services and gives a database-independent answer
+  #
+  # Returns:
+  #  The experiment status code value, and the empty string if they are not
+  require('utils')
+  require('RCurl')
+  require('rjson')
+  
+  response <- NULL
+  # username <- "username"
```
bffrost, line 2399: nit: Can we clean up this commented out line?

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -3862,7 +3928,8 @@ parseGenericData <- function(request) {
   # This is used for development: outputs the JSON rather than sending it to the
   # server and does not wrap everything in tryCatch so debug will keep printing
   developmentMode <- messenger()$devMode
-  
+  # developmentMode<-TRUE
```
bffrost, line 3931: nit: Can we clean up this commented out line?

modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -2163,7 +2170,8 @@ createNewExperiment <- function(metaData, protocol, lsTransaction, pathToGeneric
     codeType = "assay",
     codeKind = "scientist",
     lsTransaction= lsTransaction)
-  experimentStatus <- racas::applicationSettings$server.sel.experimentStatus
```
bffrost, line 2166: If I'm reading this correctly, the prior behavior was to set experimentStatus to the value of that `server.sel.experimentStatus` config. If that was null, then we'd fall back to "approved".

With your new optional field, we'd now take the user-supplied status from the SEL sheet if it's provided. But in the event it's not provided, I think this change removes the behavior of defaulting to `server.sel.experimentStatus` and would just _always_ default to `approved`.

I think it'd be better to have something like this, so the order of precedence is "value for SEL header > experimentStatus config > 'approved'":
```
  experimentStatus <- metaData$"Experiment Status"
  if (is.null(experimentStatus) || experimentStatus == "") {
    experimentStatus <- racas::applicationSettings$server.sel.experimentStatus
  }
  if (is.null(experimentStatus) || experimentStatus == "") {
    experimentStatus <- "approved"
  }
```

bffrost (changes requested):

modules/GenericDataParser/src/server/generic_data_parser.R
philippcheung: I've implemented your recommendation.  Good catch.

modules/GenericDataParser/src/server/generic_data_parser.R
philippcheung: Removed

modules/GenericDataParser/src/server/generic_data_parser.R
philippcheung: Removed

bffrost (approved):
LGTM, neat new feature!
Bonus points for adding an acasclient test for it, so we get automated test coverage if it breaks

PR #1112, opened 2023-09-18
ACAS-702: Include analysis group values parameter for faster depedency checks

brianbolt:
## Description
 - As part of ACAS-307 (https://github.com/mcneilco/acas/pull/1079) we started including analysis group values when fetching lot dependencies which can be very slow if there are a lot of experimental results.
 - The lot details page check lot dependencies when loading the page and therefore the loading of some lot details pages slowed significantly with ACAS-307.
 - This change adds a "include_analysis_group_values" parameter to the python ACAS client `get_lot_dependencies` method and defaults it true for backwards compatibility.
 - This default for the actual acas route is false so the lot details page and any other page which loads dependencies of the lots sees a significant increase in speed.

## Related Issue
ACAS-702

## How Has This Been Tested?
Ran acas client tests
Manually loaded a large 50K line experiment all to the same lot.  Witnessed the very slow lot details page load before the code change.  Applied the code change and verified the lot details page loaded almost instantly.


iamanandkr (approved):
LGTM!

PR #1113, opened 2023-09-18
ACAS-700: Make chemist, notebook page and synthesis date editable in the UI by authorized users after lot save.

brianbolt:
## Description
 - Allows editing of notebook, synthesis date and notebook page fields by users who are able to edit lots.  This is something that is already by the API but restricted via the GUI.

## Related Issue
ACAS-700

## How Has This Been Tested?
Created a lot via the GUI and verified these fields are all still required by the GUI when creating new.
Edited the lot and verified:
  - The fields are required still when editing lot
  - I could edit the lot chemist, synthesis and notebook pages now via the GUI
  - I could not edit these fields as a user a cmpdreg user which did not have access to edit the lot.

### DIFF
--- a/modules/CmpdReg/src/client/custom/LotView_Custom.inc
+++ b/modules/CmpdReg/src/client/custom/LotView_Custom.inc
@@ -5,11 +5,11 @@
 		<div>
 			<div>
 				<label class="FormLabel">*Chemist:</label>
-				<select class="FormInput chemist" <%- saved ? 'disabled=true': '' %> ></select>
+				<select class="FormInput chemist" ></select>
 			</div>
 			<div>
 				<label class="FormLabel"><span class="notForVirtual">*</span>Notebook-Page:</label>
-				<input type="textfield" class="FormInput notebookPage" <%- saved ? 'disabled=true': '' %> value="<%- notebookPage %>" />
+				<input type="textfield" class="FormInput notebookPage" value="<%- notebookPage %>" />
 			</div>
 		</div>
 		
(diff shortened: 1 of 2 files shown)


modules/CmpdReg/src/client/custom/Lot_Custom.js
```diff
@@ -524,6 +522,8 @@ $(function () {
         if (this.vendorCodeController.getSelectedModel().isNew()) vendor = null;
         else vendor = this.vendorCodeController.getSelectedModel();
         this.model.set({
+          notebookPage: jQuery.trim(this.$('.notebookPage').val()),
+          synthesisDate: jQuery.trim(this.$('.synthesisDate').val()),
```
brianbolt, line 526: Chemist was set below already so just needed to add these lines to get the notebook page and synthesis date. always.

modules/CmpdReg/src/client/custom/LotView_Custom.inc
```diff
@@ -70,7 +70,7 @@
 			</div>
 			<div>
 				<label class="FormLabel"><span class="notForVirtual">*</span>Synthesis Date:</label>
-				<input type="textfield" class="FormInput synthesisDate" <%- saved ? 'disabled=true': '' %> value="<%- synthesisDate %>"/>
+				<input type="textfield" class="FormInput synthesisDate" value="<%- synthesisDate %>"/>
```
brianbolt, line 73: This line and the 2 above just remove the logic that said to disable the field if it was saved.  Instead, it's always editable as long as the lot itself is editable.

iamanandkr (approved):
LGTM! Thank you for the comments. It helped a lot in following the changes 🙏

PR #1114, opened 2023-09-19
ACAS-701: Return status code 404 for get metalot route

brianbolt:
## Description
 - ACAS should return 404, not 500 when a lot does not exist.  Currently it returns 500 and on certain nginx proxied servers, the 500 error is swallowed and the front end cannot check for the error as it's replaced with an html structure.  Then end result is that the UI hangs when a lot does not exist instead of showing "Lot does not exist" on the page.

## Related Issue
ACAS-702

## How Has This Been Tested?
Ran manual test in the UI to verify that the correct message is displayed.
Ran ACAS client tests and verified tests pass.


iamanandkr (approved):
LGTM! I see we have a number of other usages where the status code is set to 500, should we create a separate case?

brianbolt:
> LGTM! I see we have a number of other usages where the status code is set to 500, should we create a separate case?

Yes, thanks that's probably a good idea.  I'll do that later today.

PR #1119, opened 2023-10-18
ACAS-716: Change tag grep to look for release tag format

bffrost:
## Description
Updated the grep used to identify latest dev tag to be more selective and only look for ####.#.#-dev## formatted tags.

## How Has This Been Tested?
Ran grep locally and on linux machine to confirm it gives the right result. Have not tested in GitHub actions as it requires this being merged.

### DIFF
--- a/.github/workflows/trigger-tag.yml
+++ b/.github/workflows/trigger-tag.yml
@@ -24,7 +24,7 @@ jobs:
       # and 2022.1.1-dev9 will become 2022.1.1-dev10
       - name: Get next dev tag name
         run: |
-          LAST_TAG=$(git tag --sort=v:refname --merged ${{ env.BRANCH_NAME }} | grep -e '-dev' | tail -1)
+          LAST_TAG=$(git tag --sort=v:refname --merged ${{ env.BRANCH_NAME }} | grep -E '^[0-9]{4}\.[0-9]+\.[0-9]+-dev[0-9]+$' | tail -1)
           echo "NEXT_TAG=$(echo $LAST_TAG | awk -F-dev -v OFS=-dev '{$NF += 1 ; print}')" >> $GITHUB_ENV
       # Trigger the build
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-tag.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       # and 2022.1.1-dev9 will become 2022.1.1-dev10
       - name: Get next dev tag name
         run: |
-          LAST_TAG=$(git tag --sort=v:refname --merged ${{ env.BRANCH_NAME }} | grep -e '-dev' | tail -1)
+          LAST_TAG=$(git tag --sort=v:refname --merged ${{ env.BRANCH_NAME }} | grep -E '^[0-9]{4}\.[0-9]+\.[0-9]+-dev[0-9]+$' | tail -1)
```
iamanandkr, line 27: Sanity Check: We are now searching for `####.######-dev######`. We use a `+` for the quarter section and so that would be one and more integer values.

iamanandkr (approved):
LGTM! I also double checked and now the tag identifier is correct on the branch.

I am curious about if we understand why it happened i.e why did `ACAS-473-dev3` even got created?
```
$ git tag --sort=v:refname --merged release/2023.2.x | grep -e '-dev' | tail 
2023.2.0-dev25
2023.2.0-dev26
2023.2.0-dev27
2023.2.2-dev0
2023.2.2-dev2
2023.2.2-dev3
2023.2.2-dev4
2023.2.2-dev5
2023.2.2-dev6
ACAS-473-dev3
```
On the already tagged commits, I don't see anything like it.

.github/workflows/trigger-tag.yml
iamanandkr: Ignore. I got confused.

brianbolt:
> I am curious about if we understand why it happened i.e why did `ACAS-473-dev3` even got created?

Yeah, not obvious to me either.  I suppose we can check deeper if we see this again.

brianbolt:
@bffrost I'm going to merge this and follow this through the build process to make sure things downstream get updated properly.

PR #1127, opened 2023-10-24
ACAS-718: Fallback LR ownership update to service user

brianbolt:
## Description
 - When creating an experiment, a user may have access to a project in ACAS but not in LiveDesign.  In this case we cannot set the owner of a LiveReport created by ACAS to the logged in user.
 - The code now attempt to update the LiveReport owner but will fall back if there is an error to the service user who created the LiveReport.

## Related Issue
ACAS-718

## How Has This Been Tested?
Updated the code on a running server and verified that fall back messages were printed and the LiveReport was properly created when the user did not have permissions in LiveDesign.


bffrost (approved):
LGTM. This will continue to update the folder and make the LR read-only, even if the user is missing permissions, but it'll gracefully catch & log if the LR owner can't be updated to the logged in user.

PR #1130, opened 2023-10-27
ACAS-714: Add BasicThingValidateAndSaveController to required client scripts

brianbolt:
## Description
 - modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js is used as a starter template for all scripts included in ACAS sent to the client.  Upon building ACAS, any scripts that are not in the file, are included and ordered lexicographically after the scripts in this starter template. i.e. `Zmodule/script.js` is ordered after `Amodule/script.js`
 - The template was missing `Components/BasicThingValidateAndSave.js` and therefore the script was included lexicographically after some custom modules and the import order would cause any controllers which used this compnonent to fail. e.g. The final required client scripts would look like:
 
 ```
	'/javascripts/src/ACustomCompoenent/ScriptWhichExtendsBasicThingValidateAndSaveController.js',
	'/javascripts/src/Components/BasicThingValidateAndSave.js',
```
 - The workaround to this was to override the RequiredClientScripts_template.js in the acas custom repo to include basic validate and save. This change makes that step unnecessary

## Related Issue
ACAS-714

## How Has This Been Tested?
Basic tests on a custom deployment to make sure this works.


dalejerikson (approved):
👍

bffrost:
LGTM as well

PR #1132, opened 2023-11-09
ACAS-723: Revert edit parents acls

brianbolt:
## Description
 - Partially reverting #1047 until we get acls right to edit parent structures
 - No changes to acasclient were required because we commented out a number of the tests previously due to flakiness

## Related Issue
ACAS-723

## How Has This Been Tested?
 - Ran acasclient tests for edit parent
 - Verified I could login as a Creg user (non admin) and edit a parent if I could edit the lot.


bffrost (approved):
LGTM, sorry for the troubles I caused!

PR #1133, opened 2023-11-17
Revert "Revert "ACAS-493 upgrade to nodejs 18""

bffrost:
Switch ACAS back to nodejs 18.


brianbolt (approved):
LGTM. I saw your notes on the ticket about the build process for testing and that LGTM as well.

PR #1134, opened 2023-11-17
Enable Docker build caching for ACAS

bffrost:
## Description
ACAS builds take a while. It'd be nice to speed these up by caching docker layers.
I also threw in a change to the triggers of the CI workflow, as it seemed to be overeager and was triggering whenever a new branch was created and pushed, rather than only when tags (refs/tags) were pushed.

## How Has This Been Tested?
It will be tested by the GHA runs on this PR

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -61,13 +61,15 @@ jobs:
           # Check out the branch specified by ACAS_CLIENT_REF
           ref: ${{ env.ACAS_CLIENT_REF }}
       - name: Build (no push)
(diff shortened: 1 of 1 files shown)


brianbolt (approved):
LGTM

>I also threw in a change to the triggers of the CI workflow, as it seemed to be overeager and was triggering whenever a new branch was created and pushed, rather than only when tags (refs/tags) were pushed.
FWIW, it was on purpose to build and test fully on branch pushes but in practice it doesn't work very well given that the other repositories have to have matching branches already built to make this consistently work as intended.

PR #1135, opened 2023-11-28
ACAS-725: Configurable sel created protocol/experiment text default to null

brianbolt:
## Description
 - Changes the default short descriptions of protocol/experiments to null
 - Makes configurable the default short descriptions of protocols/experiments
 - This does not automatically set all protocols and experiments previously created to the default settings. This can be accomplished by running the following sqls:

```
update protocol set short_description = null where short_description like '%created by generic data parser';
update experiment set short_description = null where short_description like '%created by generic data parser';
```
## Related Issue
ACAS-725

## How Has This Been Tested?
- Verified in the GUI that the protocol and experiments both had empty descriptions by default
 
Screenshot 2023-11-28 at 10 51 24 AM
Screenshot 2023-11-28 at 10 52 35 AM
  

- Verified in the database that the protocol and experiments both had empty descriptions by default

```
acas=> select short_description from protocol where short_description is null;
 short_description 
-------------------
 
(1 row)

acas=> select short_description from experiment where short_description is null;
 short_description 
-------------------
 
(1 row)
```

- I then set the configs to the following to restore the default to their previous defaults:
```
server.sel.protocolDescription=${client.protocol.label} created by generic data parser
server.sel.experimentDescription=${client.experiment.label} created by generic data parser
```

- Verified in the GUI that the protocol and experiments both had these values set properly on creation
Screenshot 2023-11-28 at 10 56 39 AM
Screenshot 2023-11-28 at 10 56 49 AM

- Verified in the Database that the protocol and experiments both had these values set properly on creation (but didn't grab the text to post here)

- Verified that the sql in the description would set the values to null correctly and that the GUI reflected those changes.

### DIFF
--- a/conf/config.properties.example
+++ b/conf/config.properties.example
@@ -506,6 +506,12 @@ server.sel.protocolStatus=created
 server.sel.assayStage=unassigned
 server.sel.validateDoseResponseCurves=true
 
+# Default protocol/experiment description when loaded through experiment loader
+# server.sel.protocolDescription=${client.protocol.label} created by generic data parser
+# server.sel.experimentDescription=${client.experiment.label} created by generic data parser
+server.sel.protocolDescription=
+server.sel.experimentDescription=
+
 # Cron- change to "false" to not start cron jobs on startup
 server.service.cron.startOnRestart=true
 server.service.cron.enable=true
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM

PR #1144, opened 2024-02-29
ACAS-318: Delete file on Load another

brianbolt:
## Description
 - Fixes issue where file is not deleted on "Load Another" by programmatically clicking the delete button.
 - This was kind of a deep issue and there was a previous "hack" found on other modules that did this exact same thing so I went with this solution.

## Related Issue
ACAS-318

## How Has This Been Tested?
demo


bffrost (approved):
I laughed out loud at what the fix is. LGTM

PR #1146, opened 2024-03-28
ACAS-761: Fix download urls for experiment files containing # special character

bffrost:
## Description
I dug into this a little more and found that there are some subtle differences between `encodeURI` which is meant to encode an entire URL / URI and `encodeURIComponent` which is meant to encode a query string or path component in a URL. For more info: https://stackoverflow.com/questions/75980/when-are-you-supposed-to-use-escape-instead-of-encodeuri-encodeuricomponent

I found that a simple switch to encoding only the `fileValue` with `encodeURIComponent` fixes the issue with `#` characters.

## Related Issue
https://schrodinger.atlassian.net/browse/ACAS-761

## How Has This Been Tested?
- Tested uploading an experiment file whose filename contained `#` and confirmed I could download it again
- Also tested a horrible file with name: `Generic Special Chars ! @ # $ % ^ & * ( ) - _ = + [ ] { } | \ / < > ' " ? ,.xlsx`. This filename gets clobbered during the upload, and gets saved to the ACAS filestore as ` : < > ' %22 ? ,.xlsx`, which is a related but separate bug. Before this fix, that file could not be downloaded from the experiment browser, and after the fix it can be downloaded successfully.

### DIFF
--- a/modules/ServerAPI/src/client/AttachFile.coffee
+++ b/modules/ServerAPI/src/client/AttachFile.coffee
@@ -102,14 +102,14 @@ class BasicFileController extends AbstractFormController
 		$(@el).empty()
 		$(@el).html @template(@model.attributes)
 		fileValue = @model.escape('fileValue')
-		urlValue = @model.excape('urlValue')
+		urlValue = @model.escape('urlValue')
 		if ((fileValue is null or fileValue is "" or fileValue is undefined) and (urlValue is null or urlValue is "" or urlValue is undefined))
 			@createNewFileChooser()
 		else
 			if urlValue?
 				@$('.bv_uploadFile').html '<div style="margin-top:5px;margin-left:4px;"> <a href="'+ encodeURI(urlValue)+'">'+urlValue+'</a></div>'
 			else
-				@$('.bv_uploadFile').html '<div style="margin-top:5px;margin-left:4px;"> <a href="'+encodeURI(window.conf.datafiles.downloadurl.prefix+fileValue)+'">'+@model.escape('comments')+'</a></div>'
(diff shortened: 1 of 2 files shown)


modules/ServerAPI/src/client/AttachFile.coffee
```diff
@@ -102,14 +102,14 @@ class BasicFileController extends AbstractFormController
 		$(@el).empty()
 		$(@el).html @template(@model.attributes)
 		fileValue = @model.escape('fileValue')
-		urlValue = @model.excape('urlValue')
+		urlValue = @model.escape('urlValue')
```
bffrost, line 105: @brianbolt I believe this is purely a typo so I've corrected it without testing

modules/ServerAPI/src/client/AttachFile.coffee
```diff
@@ -102,14 +102,14 @@ class BasicFileController extends AbstractFormController
...
 		if ((fileValue is null or fileValue is "" or fileValue is undefined) and (urlValue is null or urlValue is "" or urlValue is undefined))
 			@createNewFileChooser()
 		else
 			if urlValue?
 				@$('.bv_uploadFile').html '<div style="margin-top:5px;margin-left:4px;"> <a href="'+ encodeURI(urlValue)+'">'+urlValue+'</a></div>'
```
bffrost, line 110: Here I intentionally left `encodeURI` because the urlValue may be an entire URL

modules/ServerAPI/src/client/AttachFile.coffee
brianbolt: Whoops, good catch. I agree with the change.

brianbolt (approved):
LGTM.  

Just pointing out additionally that any ACAS data viewers like LiveDesign would also need employ its own mechanism of escaping the path as it is actually stored in the DB with `#` character.

PR #1148, opened 2024-04-03
ACAS-710: Code value picklists type ahead search

brianbolt:
## Description
 - See related acas-roo-server changes https://github.com/mcneilco/acas/pull/1148
 - Updated code table acas form widget to type ahead search rather than pre-fetch all picklist items.
 - In the case that there is a selected value, it will fetch the selected value only so that the intial value will be set. 

## Related Issue
ACAS-710

## How Has This Been Tested?
- Ran acasclient tests to verify compatability.
- Uploaded thousands of data column/ column tree ls_type / ls_kind values to ddict value and verified the code table fields behaved properly in Experiment editor and protocol editor as well as those in each of the endpoint manage fields.

Demo video
demo

### DIFF
--- a/modules/CodeTablesAdmin/src/server/routes/CodeTablesAdminRoutes.coffee
+++ b/modules/CodeTablesAdmin/src/server/routes/CodeTablesAdminRoutes.coffee
@@ -41,7 +41,7 @@ exports.validateCodeTablesEntityBeforeSave = (req, resp) ->
 	config = require '../conf/compiled/conf.js'
 	codeTableServiceRoutes = require "./CodeTableServiceRoutes.js"
 	searchTerm = req.params.searchTerm
-	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, (results) ->
+	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, null, (results) ->
 		if !req.body.data?
 			data = req.body
 		else
@@ -105,15 +105,7 @@ exports.searchCodeTablesEntities = (req, resp) ->
 	request = require 'request'
 	config = require '../conf/compiled/conf.js'
 	codeTableServiceRoutes = require "./CodeTableServiceRoutes.js"
-	searchTerm = req.params.searchTerm.toLowerCase().trim()
-	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, (results) ->
-		searchResults = []
-		if searchTerm == "*" || searchTerm == ""
-			searchResults = results
-		else
-			for r in results
-				if (r.code? && r.code.toLowerCase().indexOf(searchTerm) >= 0) || (r.comments? && r.comments.toLowerCase().indexOf(searchTerm) >= 0) || (r.description? && r.description.toLowerCase().indexOf(searchTerm) >= 0) || (r.name? && r.name.toLowerCase().indexOf(searchTerm) >= 0) 
-					searchResults.push(r)
+	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, searchTerm, (results) ->
 		resp.json searchResults
 
 exports.getCodeTablesEntities = (req, resp) ->
--- a/modules/Components/src/client/ACASFormFields.coffee
+++ b/modules/Components/src/client/ACASFormFields.coffee
@@ -399,9 +399,9 @@ class ACASFormLSCodeValueFieldController extends ACASFormAbstractFieldController
 			else
 				## Editable picklists can type ahead serch code tables
 				## we want to first populate with the matching value only by adding a shortName parameter
-				@pickList.url = "/api/codetables/#{mdl.get 'codeType'}/#{mdl.get 'codeKind'}/"
+				@pickList.url = "/api/codetables/#{mdl.get 'codeType'}/#{mdl.get 'codeKind'}/?maxHits=100"
 				if mdl.get('value')?
-					@pickList.url = "#{@pickList.url}?shortName=#{mdl.get('value')}"
+					@pickList.url = "#{@pickList.url}&shortName=#{mdl.get('value')}"
 			plOptions =
 				el: @$('.bv_editablePicklist')
 				collection: @pickList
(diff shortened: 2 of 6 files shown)


modules/Components/src/client/ACASFormFields.coffee
```diff
@@ -435,6 +439,9 @@ class ACASFormLSCodeValueFieldController extends ACASFormAbstractFieldController
 		if @options.autoSavePickListItem? && !@options.autoSavePickListItem
 			plOptions.autoSave = false
 
+		plOptions.autoFetch = true
+		if @options.autoFetch?
+			plOptions.autoFetch = @options.autoFetch
```
brianbolt, line 444: *sort of outside pr*: Previously there was no way to pass auto fetch to the picklist controller.  At one point I thought I could turn of auto fetching of the picklists but forgot about the "selectedCode" use case where the picklist actually needs to get populated with the selected value so it needs atleast that item int he picklist. 

TLDR; I left this in as a feature but it defaults to true like before.

modules/Components/src/client/PickList.coffee
```diff
@@ -361,7 +361,28 @@ class PickListSelect2Controller extends PickListSelectController
 			openOnEnter: false
 			allowClear: true
 			width: @width
-
+			ajax:
```
brianbolt, line 364: This is most of the magic.  The select2 widget lets you add ajax calls for type ahead search.  I'm leveraging that.  I had to do this sort of hacky thing create a real URL to remove the "shortName" parameter if it existed in the list and only include the labelTextSearch and maxHits parameters.

modules/Components/src/client/PickList.coffee
```diff
@@ -685,6 +706,7 @@ class EditablePickListSelect2Controller extends EditablePickListSelectController
 			collection: @collection
 			selectedCode: @options.selectedCode
 			filters: filters
+			autoFetch: @options.autoFetch
```
brianbolt, line 709: *sort of outside pr scope*: Pass options auto fetch to the picklist controller options as noted elsewhere in my review.

modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
```diff
@@ -37,31 +37,33 @@ exports.getAllCodeTableValues = (req, resp) ->
 		)
 
 exports.getCodeTableValues = (req, resp) ->
-	exports.getCodeTableValuesInternal req.params.type, req.params.kind, (result) ->
+	exports.getCodeTableValuesInternal req.params.type, req.params.kind, req.query, (result) ->
```
brianbolt, line 40: This and the few lines below are the main changes on the acas side.  I added backwards compatability for the old syntax so out of date customer specific server functions would work the same.  This passes the query parameters directly to roo.

modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
```diff
@@ -37,31 +37,33 @@ exports.getAllCodeTableValues = (req, resp) ->
...
+			qs = query
 		request = require 'request'
 		request(
 			method: 'GET'
 			url: baseurl
+			qs: qs
 			json: true
 		, (error, response, json) =>
 			if !error && response.statusCode == 200
-				if json.length > 0 and !json[0].displayOrder? and json[0].name?
```
brianbolt, line 58: I don't know why this was added. Roo always did this same sorting before passing the response and it was messing up my "best match" sort when searching by labelTextSerch so I removed it.

modules/ServerAPI/src/server/CustomerSpecificServerFunctions.coffee
```diff
@@ -409,7 +409,7 @@ exports.getAllAuthors = (opts, callback) ->
 		# This is was added for the purpose of allowing additional non-authors to show up in picklists throughout ACAS and Creg
 		if opts.additionalCodeType? and opts.additionalCodeKind?
 			codeTableServiceRoutes = require "#{ACAS_HOME}/routes/CodeTableServiceRoutes.js"
-			codeTableServiceRoutes.getCodeTableValuesInternal opts.additionalCodeType, opts.additionalCodeKind, (codes) ->
+			codeTableServiceRoutes.getCodeTableValuesInternal opts.additionalCodeType, opts.additionalCodeKind, null, (codes) ->
```
brianbolt, line 412: Everything below here is just to match the new codeTableValuesInternal syntax

modules/Components/src/client/PickList.coffee
bffrost: Indeed this is kinda hacky but it's reasonable. You're programmatically getting the base path of the URL, so this looks reliable.

modules/CodeTablesAdmin/src/server/routes/CodeTablesAdminRoutes.coffee
```diff
@@ -105,15 +105,7 @@ exports.searchCodeTablesEntities = (req, resp) ->
...
-	searchTerm = req.params.searchTerm.toLowerCase().trim()
-	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, (results) ->
-		searchResults = []
-		if searchTerm == "*" || searchTerm == ""
-			searchResults = results
-		else
-			for r in results
-				if (r.code? && r.code.toLowerCase().indexOf(searchTerm) >= 0) || (r.comments? && r.comments.toLowerCase().indexOf(searchTerm) >= 0) || (r.description? && r.description.toLowerCase().indexOf(searchTerm) >= 0) || (r.name? && r.name.toLowerCase().indexOf(searchTerm) >= 0) 
-					searchResults.push(r)
+	codeTableServiceRoutes.getCodeTableValuesInternal req.params.codeType, req.params.codeKind, searchTerm, (results) ->
```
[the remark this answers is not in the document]
bffrost: Ah, I see. This was an older node-server implementation of searching codeTables that we think was unused, but you've now ported it to use the database-based full text search.
I'm in favor of this change, and makes sense that we can tolerate the slight search behavior change given we don't think `searchCodeTablesEntities` is actually being used anywhere.

bffrost (approved):
LGTM. I appreciate the thought you put into backward compatibility and some of the minor cleanup you've done along the way.

modules/Components/src/client/PickList.coffee
brianbolt: As a side note, this might be a way of fixing our relative vs full path urls problem with ACAS sometime int he future:
https://stackoverflow.com/questions/16553026/prepend-absolute-url-path-in-backbone-sync-for-an-api-on-another-server

PR #1152, opened 2024-04-16
ACAS-764: Clean up cmpdreg exported search results > 24 hours old

brianbolt:
## Description
 - Delete exported search results that are > 24 hours since last modified after the export searchResults route is called

## Related Issue
ACAS-764

## How Has This Been Tested?
1. Basic
Exported some search results and then ran: `touch -d "25 hours ago" privateUploads/exportedSearchResults/*.sdf`
Verified the files were cleaned up

2. Folder
Added a folder to exportedSearchResults and verified this did not produce and error and was skipped

3. Added a purposeful error inside the `cleanUpExportedSearchResults` function and verified that the exportSearchResults function continue to work (just to make sure this doesn't break export if it happens to error for some reason.

4. Deployed this to a stage server with a lot of old exported search results and verified it cleaned up old files when I exported search results.

### DIFF
--- a/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
+++ b/modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
@@ -927,6 +927,7 @@ exports.exportSearchResults = (req, resp) ->
 	serverUtilityFunctions.ensureExists exportedSearchResults, 0o0744, (err) ->
 		if err?
 			console.log "Can't find or create exportedSearchResults folder: " + exportedSearchResults
+			console.error err
 			resp.statusCode = 500
 			resp.end "Error trying to export search results to sdf: Can't find or create exportedSearchResults folder " + exportedSearchResults
 		else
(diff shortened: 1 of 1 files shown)


modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
```diff
@@ -927,6 +927,7 @@ exports.exportSearchResults = (req, resp) ->
 	serverUtilityFunctions.ensureExists exportedSearchResults, 0o0744, (err) ->
 		if err?
 			console.log "Can't find or create exportedSearchResults folder: " + exportedSearchResults
+			console.error err
```
brianbolt, line 930: *outside pr scope*: I wanted this to print the actual error. I initial had a weird local setup that was producing and an error and I added this line to debug the issue.

modules/CmpdReg/src/server/routes/CmpdRegRoutes.coffee
```diff
@@ -967,6 +968,47 @@ exports.exportSearchResults = (req, resp) ->
 
 			)
 
+	# After responding, clean up old exported search results
+	exports.cleanUpExportedSearchResults(exportedSearchResults)
```
brianbolt, line 972: This occurs after the resp.end calls so won't interfere with the request.

bffrost (approved):
LGTM. TIL we can run more code after resp.end, but makes sense since it's all async and not a true "return"

PR #1155, opened 2024-04-18
ACAS-727: Update to nodejs 20 LTS

brianbolt:
## Description
 - Bumped nodejs from 18.x to 20.x
 - See equivalent changes for races and acas-roo-server
   - https://github.com/mcneilco/racas/pull/92
   - https://github.com/mcneilco/acas-roo-server/pull/463

## Related Issue
ACAS-727

## How Has This Been Tested?
Ran acasclient tests after building all 3 branches of ACAS-727 locally
Poked around the UI and loaded all module menus, verified drop downs worked.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ RUN   dnf install -y initscripts python3-psycopg2
 
 # node
 ENV NPM_CONFIG_LOGLEVEL warn
-ENV NODE_VERSION 18.x
+ENV NODE_VERSION 20.x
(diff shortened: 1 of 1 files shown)


dalejerikson (approved):

bffrost (approved):
Wow, great that this is so easy. I checked on the ACAS-762 issue mentioned below but it seems node 20 doesn't yet have the patched version of npm, so I think the workaround needs to stay for a little longer.

PR #1161, opened 2024-05-24
ACAS-739: Open in LiveDesign fails fails for image uploads experiments

tanay-makharia:
## Description
Removed the prefix 'inlineFileValue_' from assays_to_add -> resultType

## Related Issue
[ACAS-739](https://schrodinger.atlassian.net/browse/ACAS-739)

## How Has This Been Tested?
Directly edited the code on cluster and the error is no longer on the log stream

### DIFF
--- a/modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
+++ b/modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
@@ -372,6 +372,11 @@ def main():
 
     compound_ids=args['input']['compounds']
     assays_to_add=args['input']['assays']
+
+    # Remove the inlineFileValue_ prefix from the resultType
+    for assay_to_add in assays_to_add:
+        assay_to_add['resultType'] = re.sub(r'^inlineFileValue_', '', assay_to_add['resultType'])
+
     experiment_code=args['input']['experimentCode']
     try:
         project=args['input']['project']
--- a/modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
+++ b/modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py
@@ -177,12 +177,25 @@ def wait_and_add_columns(ld_client,
     :param delay_sec: Seconds of delay been tries, defaults to 5
     """
     success = False
+
+    pattern_match_found = False
+    for col_name in column_names:
+        # check if column name with prefix '{assay_name} (inlineFileValue_' exists
+        # taking this pattern from get_assay_names_to_full_col_names -> col_name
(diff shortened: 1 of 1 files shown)


bffrost (changes requested):
@tanay-makharia Thanks for posting this fix. I think it will work for any systems that have been migrated to the new DI mappings (where `inlineFileValue_` is no longer present) but I am worried this will break systems that still have old DI mappings (where `inlineFileValue_` is still present). I mentioned in a comment on the Jira case that we should think about making this backward compatible by checking for both "flavors" of the assay name. Can you come up with some ideas about how we might be able to make this script backward compatible so it works with either version of the DI mappings?

brianbolt (changes requested):
Frost mentioned in the PR that it's possible that not everyone will want or use the new DI mappings which remove inlineFileValue from the DI query.  The built in view which is what is being used to get the result types in `assays_to_add` is [adding `inlineFileValue`](https://github.com/mcneilco/acas-roo-server/blob/247ca02aaf1a984a915916d5f4e9e29fe859ce60/src/main/resources/db/migration/postgres/V1.0.6.8__restore_api_views.sql#L66) the the resultType you see here.  I do like the approach to just fix this in the python script (as opposed to removing inline file value from the view) but I'd like to see us fix it in a backwards compatible way.

Suggestion from me would be to 
- append change this function to have it detect "inlineFileValue_" result kinds and append the none inline file value to the list.
- Update how [this function](https://github.com/mcneilco/acas/blob/7fa35b489cbbcdb9cc0137169b05b6b06aa2654a/modules/ServerAPI/src/server/python/createLiveDesignLiveReportForACAS/create_lr_for_acas.py#L184) works by having it expect that either the inlinefilevalue version or the non-inlinefilevalue will be added to the LR which may take a bigger refactor of the code.

brianbolt:
> @tanay-makharia Thanks for posting this fix. I think it will work for any systems that have been migrated to the new DI mappings (where `inlineFileValue_` is no longer present) but I am worried this will break systems that still have old DI mappings (where `inlineFileValue_` is still present). I mentioned in a comment on the Jira case that we should think about making this backward compatible by checking for both "flavors" of the assay name. Can you come up with some ideas about how we might be able to make this script backward compatible so it works with either version of the DI mappings?

I think @bffrost and I were basically saying the same thing.  I gave a partial suggestion on how you might achieve this.

tanay-makharia:
I tested this on new mapping and no error were logged.
I updated the mapping view query to previous mapping and no error were logged when I added the data to LR

bffrost (approved):
LGTM

dalejerikson (approved):
LGTM!

brianbolt (approved):
LGTM. Thanks for the work on this.

tanay-makharia:
Closing this as merging to a different release(release/2024.1.x)
New PR - https://github.com/mcneilco/acas/pull/1162

PR #1165, opened 2024-06-05
ACAS-774: Support R-4.4.0 changes

brianbolt:
## Description
 - See related base R image changes here: https://github.com/mcneilco/racas/pull/93
 - This pull request includes changes to support R-4.4.0. The changes involve updating the LD_LIBRARY_PATH and PATH environment variables based on the architecture of the system. Additionally, there are updates to the getExperimentByNameCheck and runMain functions to handle NA values correctly following:

https://cran.r-project.org/bin/windows/base/old/4.2.0/NEWS.R-4.2.0.html
> Calling && or || with either argument of length greater than one now gives a warning (which it is intended will become an error).

In 4.3.0 this became an error: https://cran.r-project.org/bin/windows/base/old/4.3.0/NEWS.R-4.3.0.html
> Calling && or || with LHS or (if evaluated) RHS of length greater than one is now always an error, with a report of the form
    'length = 4' in coercion to 'logical(1)'

So now we get:

https://github.com/mcneilco/acas/blob/9d261eefd5600d3d650bc707c1476b788e346a6e/modules/GenericDataParser/src/server/generic_data_parser.R#L2008
```
Error in is.na(protocol) || protocolOfExperiment$id != protocol$id : 
  'length = 15' in coercion to 'logical(1)'
```

## Related Issue
ACAS-744

## How Has This Been Tested?
 - Ran acasclient tests
 - Ran manual load of dose response data and verified curves rendered and results were expected in the UI.  Ran curve curator and verified it functioned properly (this is some of the more complex R code we run).

### DIFF
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -4631,7 +4631,6 @@ validateSubjectData <- function(subjectData, dryRun) {
   if (is.null(subjectData)) {
     return(NULL)
   }
-  # subjectData is just list() with the new thing so why is it 
   uniqueDF <- unique(subjectData[, c("Class", "valueKind")])
   uniqueDF <- uniqueDF[!(uniqueDF$valueKind %in% c('flag cause', 'flag observation', 'flag status')), ]
   validateValueKinds(uniqueDF$valueKind, uniqueDF$Class, dryRun, reserved = NULL)
--- a/modules/GenericDataParser/src/server/generic_data_parser.R
+++ b/modules/GenericDataParser/src/server/generic_data_parser.R
@@ -2005,7 +2005,7 @@ getExperimentByNameCheck <- function(experimentName, protocol, configList, dupli
     protocolOfExperiment <- getProtocolByCodeName(experimentList[[1]]$protocol$codeName)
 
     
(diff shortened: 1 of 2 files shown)


modules/GenericDataParser/src/server/generic_data_parser.R
```diff
@@ -4628,6 +4628,10 @@ getSubjectAndTreatmentData <- function (precise, genericDataFileDataFrame, calcu
 validateSubjectData <- function(subjectData, dryRun) {
   # Validates Subject Data
   # For now, just passes information to validateValue Kinds
+  if (is.null(subjectData)) {
+    return(NULL)
+  }
+  # subjectData is just list() with the new thing so why is it
```
bffrost, line 4634: @brianbolt this comment looks incomplete. Is this intentional?

bffrost (approved):
`setenv.sh` and `is.na` -> `identical` changes LGTM

One code comment looks kindof accidental so I've questioned it.

modules/GenericDataParser/src/server/generic_data_parser.R
brianbolt: I did not mean to leave this comment in, thanks for noting it.  I removed it.

PR #1168, opened 2024-06-05
ACAS-762-revert

brianbolt:
## Description
 - Removes pinned npm version now that 18.20.1 has been released https://nodejs.org/en/blog/release/v18.20.1

## Related Issue
ACAS-762

## How Has This Been Tested?
TBD on merge since we don't do mule-arch builds on PRs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,8 +37,6 @@ ENV NPM_CONFIG_LOGLEVEL warn
 ENV NODE_VERSION 18.x
 RUN curl -fsSL https://rpm.nodesource.com/setup_$NODE_VERSION | bash - && \
   dnf install -y nodejs
-# ACAS-762 temporary fix for npm multi-arch build issue. Remove when Node is updated to > 18.20.1 and fix is confirmed
-RUN npm install npm@10.5.1 -g
 
 # ACAS
 RUN	    useradd -u 1000 -ms /bin/bash runner


bffrost:
@brianbolt I'm not sure that the version `18.20.1` that I put in that comment is actually the build this fix landed in. Looking through some of the links / history from the ticket, I found this commit in node where the npm version bump landed: https://github.com/nodejs/node/commit/ad86a12964814633acc5c374feed09291a51e743

It says it's only in node `v22`

bffrost (changes requested):
Based on my comment above, I don't think we should actually merge this right now unless there's some new information. We could update the comment to be more accurate though

bffrost:
@brianbolt what's the context of the re-review request? Re-reading my earlier comments, it looks like we'd want to wait until we're using node v22+ and our Dockerfile still says node v20.x

brianbolt:
Sorry, was lining up a comment to go with my re-request but realized my testing is not complete.  Seeing this build issue currently on 2025.1.x https://github.com/mcneilco/acas/actions/runs/12678771851/job/35339632790#step:23:1279

Thought merging this would fix it but realized that the latest build of this ACAS-762-revert branch didn't actually run the multi-arch build where I'm running into the error so was going to test on a another machine.

brianbolt:
@bffrost I think this is fixed in [Node 20.13.0](https://nodejs.org/en/blog/release/v20.13.0) by upgrading to[ 10.5.1 ](https://docs.npmjs.com/cli/v10/using-npm/changelog#bug-fixes-9) which notes that that version has

[8cab136](https://github.com/npm/cli/commit/8cab136f731c69be079be08d79e3514e01bbd563) [#7324](https://github.com/npm/cli/pull/7324) ensure maxSockets is respected (#7324) (@lukekarrys)

Which is the fix for the original issue we found for the issue here https://github.com/npm/cli/issues/7072

bffrost (approved):
LGTM, makes sense based on your note that node 20.13.0 has npm upgraded to 10.5.2

PR #1166, opened 2024-06-05
ACAS-744: Switch to vault.centos.org as a workaround for stream8 EOL

brianbolt:
## Description
 - Switches all references from mirror to vault
 - I plan on up merging his PR and creating a seperate 

## Related Issue
ACAS-744

## How Has This Been Tested?
TBD

PR #1173, opened 2024-06-10
ACAS-744: Switch to vault.centos.org as a workaround for stream8 EOL

bffrost:
## Description
Backporting patch from #1166 to earlier release branch


brianbolt (approved):
LGTM

PR #1178, opened 2024-06-17
ACAS-734: Scope experiment table controller to this

brianbolt:
## Description
 - Fixes bug where clicking on rows in `Experiment Browser` an then rows in `Protocol Browser` with endpoint manager turned on would produce and alert `Cannot reinitialise DataTable.`
 
Screenshot 2024-06-17 at 4 10 44 PM

## Related Issue
ACAS-734

## How Has This Been Tested?
 - Reliably reproduced the problem by:
    - Searching for an experiment in Experiment Browser and clicking a row
    - Searching for a protocol in Protocol Browser and clicking a row
- Verified the alert went away using the same test
- Verified manually that both `Protocol Browser` and `Experiment Browser` functioned normally after the change.


bffrost (approved):
LGTM

PR #1189, opened 2024-10-01
ACAS-802: Add --no-deps --disable-pip-version-check to bootstrap ldclient script

brianbolt:
## Description
 - Without internet, pip can't reach out to check dependencies so it fails.  We are now installing ldclient dependencies into the image before this step so there is no reason to check for dependencies.
 - Also skip checking for pip upgrades which slow down startup when there is no internet with `WARNING: There was an error checking the latest version of pip.` (we don't upgrade pip anyway so its a pointless check).

## Related Issue
ACAS-802

## How Has This Been Tested?
Ran the following command on an kubernetes cluster without internet and verified ldclient installed and functioned properly:

```
kubectl exec -n livedesign deployments/acas -c acas -it -- pip install --disable-pip-version-check --upgrade --force-reinstall --no-deps --user http://livedesign-internal/livedesign/ldclient.tar.gz
```


bffrost (approved):
LGTM

PR #1193, opened 2024-11-27
ACAS-818: Jquery upgrades

brianbolt:
## Description
- Added 2 commits to ACAS-818-jquery-query upgrade and submitting this PR to get it merged to main feature branch
  - Removed old jquery-ui and fixed issues with new jquery-ui import: https://github.com/mcneilco/acas/commit/6ce625167838cfed1e7d39769cc47abcd39b41e6
  - Updated to jquery 3.7.1: https://github.com/mcneilco/acas/commit/2667e845b001358a01a275a4771dd4b2c757d6f2

## Related Issue
ACAS-818

## How Has This Been Tested?
Clicked around the UI modules


modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
```diff
@@ -1,21 +1,22 @@
 exports.requiredScripts = [
 	'/lib/jquery.min.js',
+	'/lib/jquery-migrate-1.4.1.js', // Remove after upgrading
```
bffrost, line 3: We can remove this, right?

Did you end up installing jquery-migrate 3.x and working through the warnings it showed also?

bffrost (changes requested):

modules/BuildUtilities/src/server/routes/RequiredClientScripts_template.js
brianbolt: I did install it, I fixed a few things but chose not to remove it as I figured we had some more testing to do.  I can remove it before merging if you like. What do you think?

bffrost: Let's remove it here. jquery-migrate 1.x doesn't work with jquery 3.x which this branch already gets us to. Makes sense to do some more testing after merging this, but we should use jquery-migrate 3.x anyway, and I think you and I can do that without checking in that library.

PR #1194, opened 2024-11-27
Acas 818 backbone upgrade

brianbolt:
## Description
 - To be reviewed after https://github.com/mcneilco/acas/pull/1193
 - Updates backbone in ACAS to latest 1.6.0
   - The only real fix here was to make sure options was attached to all Backbone Views as this was removed from backbone in 1.1.0 (2013)

## Related Issue
ACAS-818

## How Has This Been Tested?
Clicked around the UI looking for any errors and systematically searched for initialize methods without options.


bffrost (approved):
LGTM aside from needing to remove jquery migrate which I noted on the other PR. I'm good to merge these down to the feature branch but then I'd like to do a bit of testing on that branch to confirm things are good in my hands too

PR #1196, opened 2024-12-05
ACAS-824

brianbolt:
## Description
 - Fixes ACASLSThingBrowser and Standardization table filtering after Datatable upgrade.
 - Fixes Standardization table column visibility UI 

## Related Issue
ACAS-824

## How Has This Been Tested?
Thing browser

https://github.com/user-attachments/assets/13b03d70-df3d-4a5f-ae57-ffc65074d89e

Standardization - This one is hard to test because Indigo standardization table only works with one compound registered for some reason.  Surprised it worked at all since we never tested the non-implemented indigo standardization.

https://github.com/user-attachments/assets/1a06a20a-4ff0-4a21-ab07-7eecdf162c4d

### DIFF
--- a/modules/Components/src/client/ACASThingBrowserView.html
+++ b/modules/Components/src/client/ACASThingBrowserView.html
@@ -33,9 +33,9 @@
 <script type="text/template" id="ThingSummaryTableView" xmlns="http://www.w3.org/1999/html">
     <table class="table table-bordered table-striped table-hover" style="margin: 0px; padding: 0px; border: 0px; width: 100%; ">
         <thead style="">
-            <tr class="bv_colFilters">
-            </tr>
             <tr class = "bv_firstRow"> 
+            </tr>
+            <tr class="bv_colFilters">
             </tr>
         </thead>
         <tbody style="">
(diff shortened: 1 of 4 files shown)


modules/Components/src/client/ACASThingBrowserView.html
```diff
@@ -33,10 +33,10 @@
 <script type="text/template" id="ThingSummaryTableView" xmlns="http://www.w3.org/1999/html">
     <table class="table table-bordered table-striped table-hover" style="margin: 0px; padding: 0px; border: 0px; width: 100%; ">
         <thead style="">
-            <tr class="bv_colFilters">
-            </tr>
             <tr class = "bv_firstRow"> 
             </tr>
+            <tr class="bv_colFilters">
```
bffrost, line 38: Kindof a UX change here, to move the filter underneath the header rather than above. As long as this is intentional, I'm good with it.

bffrost (approved):
LGTM, thanks for the demos

modules/Components/src/client/ACASThingBrowserView.html
brianbolt: Ah whoops. Yes it was intentional. I forgot to mention it and also realize the standardization is not changed. I'm going to change the standardization to have the filters below the header as well.

PR #1197, opened 2025-01-03
ACAS-828 Use postgres for session management

brianbolt:
## Description
 - Switch from memory store to postgres database store for session
 - Added a new config `config.all.server.sessionTimeOutMinutes` (default 1440 = 24 hours)

## Related Issue
ACAS-823

## How Has This Been Tested?
  - Ran ACAS client tests locally
  - Basic test
     - Logged into ACAS and verified login works.
     - Restarted ACAS and verified I was still logged in when ACAS started.
     - Tried logging after a few minutes and verified it was 
     - Set the session time out 1 minute and restarted acas
     - Logged back, waited one minute and verified I was logged out automatically
 - 2 Replica test
   - Setup acas with 2 replicas in kubernetes
   - Logged into ACAS
   - Deleted the pod I was logged into and continued to work on the same window I was using previously without having to log back in.

### DIFF
--- a/app_template.coffee
+++ b/app_template.coffee
@@ -15,7 +15,7 @@ startApp = ->
 	logger = require('morgan')
 	methodOverride = require('method-override')
 	session = require('express-session')
-	MemoryStore = require('memorystore')(session)
+	PostgresqlStore = require('connect-pg-simple')(session)
 	bodyParser = require('body-parser')
 	errorHandler = require('errorhandler')
 	cookieParser = require('cookie-parser')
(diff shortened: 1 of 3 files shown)


dalejerikson (commented):
Conceptually looks good to me but the CICD is failing

brianbolt:
> Conceptually looks good to me but the CICD is failing
Yeah, thats a long standing issue where if acas-roo-server or rservices have corresponding changes, then tests fail until they are they are all merged/have been built into the main release branch builds.  In this case, I have a table being added in acas-roo-server which doesn't exist. Which will be created when this PR is merged https://github.com/mcneilco/acas-roo-server/pull/469

dalejerikson:
Gotcha, sounds good. Thanks for the explanation! :)

bffrost (commented):
~I've got the same conclusion as Dale. The changes LGTM but the CI failures don't, and they seem to be failing on login which does seem related to this change~

Just read the comments above. Makes sense.

bffrost (approved):

PR #1198, opened 2025-01-07
ACAS-818: Dependency upgrades for frontend libraries

bffrost:
## Description

Updated a number of frontend dependencies to patched or more recent versions as part of a general cybersecurity effort.
- Bootstrap to `3.4.1`
- jquery to `3.7.1`
- jqueryFileUpload to `10.32.0`
- jquery-ui-plugin to `1.14.1`
- BackboneJS to `1.6.0`
- underscore.js to `1.13.7`
- datatables to `2.1.8` 
- handsontable to last MIT licensed release (`6.6.2` from Dec 19, 2018)
- moment.js to `2.30.1`
- Removed unused ExcelApp module and its dependencies

Updated ACAS code to adapt to dependency upgrades:
- Backbone Model initialize functions are no longer automatically called with options and need to explicitly mention them
- dataTables customizations were reimplemented by @brianbolt due to upstream API changes
- underscore template function signature changed in how it requires arguments / options to be passed in

## How Has This Been Tested?

Clicked around and manually every module of ACAS in local development environment, identified and fixed several issues.

This should get a full regression pass of ACAS since the scope of the changes affect the whole UI


public/lib/jquery-migrate-1.4.1.js
bffrost, line 1: self-reminder to remove this

PR #1201, opened 2025-02-02
Revert "Revert "ACAS-828 Use postgres for session management""

brianbolt:
## Description
 - Switch from memory store to postgres database store for session
 - Added a new config `config.all.server.sessionTimeOutMinutes` (default 1440 = 24 hours)

## Related Issue
ACAS-823

## How Has This Been Tested?
  - Ran ACAS client tests locally
  - Basic test
     - Logged into ACAS and verified login works.
     - Restarted ACAS and verified I was still logged in when ACAS started.
     - Tried logging after a few minutes and verified it was 
     - Set the session time out 1 minute and restarted acas
     - Logged back, waited one minute and verified I was logged out automatically
 - 2 Replica test
   - Setup acas with 2 replicas in kubernetes
   - Logged into ACAS
   - Deleted the pod I was logged into and continued to work on the same window I was using previously without having to log back in.

### DIFF
--- a/app_template.coffee
+++ b/app_template.coffee
@@ -15,7 +15,7 @@ startApp = ->
 	logger = require('morgan')
 	methodOverride = require('method-override')
 	session = require('express-session')
-	MemoryStore = require('memorystore')(session)
+	PostgresqlStore = require('connect-pg-simple')(session)
 	bodyParser = require('body-parser')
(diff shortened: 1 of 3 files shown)


brianbolt:
Merging as this was previously approved and we decided to shift 2025.2 release https://github.com/mcneilco/acas/pull/1197

PR #1204, opened 2025-03-04
ACAS-669: Fix text overflow of long sdf property name

bffrost:
## Description

The UI looks buggy when an SDF property with a long name is uploaded. This PR fixes the CSS to overflow even if it has to break a word.

## How Has This Been Tested?

Manually in local UI:
Screenshot 2025-03-04 at 11 35 57 AM


dalejerikson (approved):
LGTM :)

PR #1205, opened 2025-03-04
ACAS-786: Add restandardization check to SaltBrowser

bffrost:
## Description

This PR adds the "allow cmpd registration" check that already exists on CReg single and bulk pages into the Salt Browser, since it also relies on chemistry services. Create, Edit, and Download buttons are disabled if compound registration is currently unavailable due to a need for a restandardization or due to unavailable chemistry services.

## How Has This Been Tested?

Tested by switching local roo to bbchem mode and pointing at a remove bbchem instance, which triggered some errors / failures from the allowCmpdRegistration route.
The table is still visible but the buttons at the top are disabled
Screenshot 2025-03-04 at 3 33 20 PM

Viewing and deleting is still available but editing is blocked
Screenshot 2025-03-04 at 3 33 25 PM


dalejerikson (approved):
LGTM

brianbolt (approved):
LGTM

PR #1207, opened 2025-03-24
Nim 1808 curve overlay coloring v2

jmcneil86043:
Description
If http request &mostRecentCurveColor is supplied as "none" or "false" then the global config option is disabled.

Sort descending by curve recordedDate is removed. However, if mostRecentCurveColor is enabled, then only that curve is moved to the top of the plot.

Because of these changes, if the request includes &plotColors aligned in order with &curveIds, that color assignment is respected

Related Issue
Tracked in a user issue tracking system

This change works with changes to racas rendering.R

How Has This Been Tested?
Stage and prod testing, then stage again when this further issue was found


brianbolt (approved):
Change LGTM but can you change the PR to merge to [release/2025.2.x](https://github.com/mcneilco/racas/tree/release/2025.2.x) instead of master?
Once thats done feel free to merge.

PR #1212, opened 2025-04-02
ACAS-859: Try shifting build args down to allow caching

bffrost:
## Description

I found that ACAS builds were not using the cache effectively, and I believe it's because of the build args which change on each build and are up near the top of the file. I'll test with multiple builds on this PR to see if that allows acas to use docker build caches effectively between builds.

## How Has This Been Tested?

Ran two builds on this PR. Saw that caching worked and sped up the Docker build time.
Also saw logging confirming that the variables I moved are being used in the configs still:
```
#23 [19/19] RUN     gulp execute:prepare_config_files
#23 0.648 working directory '/home/runner/build'
#23 0.648 setting build to: /home/runner/build
#23 0.649 setting source directories to: ["../acas","../acas_custom","../acas_shared"]
#23 0.652 [16:32:23] Using gulpfile ~/build/gulpfile.js
#23 0.652 [16:32:23] Starting 'execute:prepare_config_files'...
#23 0.694 reading configs in this order (latter configs override former configs): /home/runner/build/conf//config.properties.example
#23 0.695 reading conf file: /home/runner/build/conf//config.properties.example
#23 0.705 read conf file: /home/runner/build/conf//config.properties.example
#23 0.706 environment variable ACAS_CLIENT_ABOUT_ACAS_BUILDTIME is being substituted for config key client.about.acas.buildTime
#23 0.706 environment variable ACAS_CLIENT_ABOUT_ACAS_REVISION is being substituted for config key client.about.acas.revision
#23 0.706 environment variable ACAS_CLIENT_ABOUT_ACAS_VERSION is being substituted for config key client.about.acas.version
#23 0.743 [16:32:24] Finished 'execute:prepare_config_files' after 90 ms
```


brianbolt (approved):
LGTM

PR #1215, opened 2025-04-25
Fix _.max returning -Infinity instead of undefined when array is empty

brianbolt:


modules/ServerAPI/src/client/Label.coffee
brianbolt, line 1: Updates to this file are for 1) _.max returning -Infinity when the array is empty.  We have a lot of downstream code that checks for defined/undefined so we need to keep that behavior.

modules/ServerAPI/src/server/routes/ServerUtilityFunctions.coffee
brianbolt, line 1: Updates to this file are for 1) _.max returning -Infinity when the array is empty.  We have a lot of downstream code that checks for defined/undefined so we need to keep that behavior.

public/lib/jqueryFileUpload-10.32.0/css/jquery.fileupload-noscript.css
brianbolt, line 1: css downloaded directly from jquery file upload GitHub .

modules/ServerAPI/src/client/layout.jade
brianbolt, line 1: Looks like they split the css into 2 seperate files so replacing file upload-ui.css with the 2 I downloaded.

modules/ServerAPI/src/server/routes/ThingServiceRoutes.coffee
brianbolt, line 1: For issue 3).  The route needs to return json but roo is returning nothing if there is no error so I return just {}.  I verified the various states of the UI (duplicate project name vs not) and they both work.

dalejerikson (approved):
LGTM!

PR #1245, opened 2026-06-05
ACAS-984 Fix standardization dry-run export download parsing

brianbolt:
## Summary
Standardization dry-run result downloads could fail because the route expected a JSON response from the persistence endpoint even when that endpoint returned a plain text file path. That produced a JSON parse error and then a null response crash when reading `statusCode`.

This change updates the route calls to handle non-JSON response bodies for dry-run export/file retrieval and hardens error handling when no response object is available.

## Why this fixes it
- `standardizationDryRunSearchExport` now sends JSON request payload explicitly but does **not** force JSON parsing on the response.
- `standardizationDryRunFiles` now also treats the response as non-JSON.
- Both error paths now guard `response?.statusCode` and provide a safe fallback error body instead of throwing.

## Testing
- Reproduced the issue by running dry-run standardization and attempting to download results (observed JSON parse failure).
- Applied this patch on the server via ConfigMap override and deployment mount/restart.
- Re-tested dry-run standardization download; download now works.

<details>
<summary>Patch commands used during verification</summary>

```bash
kubectl patch configmap acas-custom-config-map \
  -n livedesign \
  --type merge \
  --patch "$(jq -Rs '{data:{\"StandardizationRoutes.coffee\": .}}' modules/Standardization/src/server/routes/StandardizationRoutes.coffee)"
```

```bash
kubectl patch deployment acas -n livedesign --type=json --patch='[
  {
    "op": "add",
    "path": "/spec/template/spec/volumes/-",
    "value": {
      "name": "standardization-routes",
      "configMap": {
        "name": "acas-custom-config-map",
        "defaultMode": 511
      }
    }
  },
  {
    "op": "add",
    "path": "/spec/template/spec/containers/0/volumeMounts/-",
    "value": {
      "name": "standardization-routes",
      "mountPath": "/home/runner/acas_custom/modules/Standardization/src/server/routes/StandardizationRoutes.coffee",
      "subPath": "StandardizationRoutes.coffee"
    }
  }
]'
```

```bash
kubectl rollout restart deployment/acas -n livedesign
kubectl rollout status deployment/acas -n livedesign
```

</details>

Then removed the patch manually later

### DIFF
--- a/modules/Standardization/src/server/routes/StandardizationRoutes.coffee
+++ b/modules/Standardization/src/server/routes/StandardizationRoutes.coffee
@@ -178,9 +178,11 @@ exports.standardizationDryRunSearchExport = (req, resp) ->
 
 	request(
 		method: 'POST'
-		body: req.body
+		body: JSON.stringify(req.body)
 		url: url
-		json: true
+		headers:
+			'Content-Type': 'application/json'
+		json: false
 		timeout: 86400000
 	, (error, response, json) =>
 		console.log "standardizationDryRunFiles response" + JSON.stringify(response)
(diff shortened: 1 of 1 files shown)


bffrost (approved):
LGTM. This was introduced by my requestAdapter implementation (elimination of requests, switching to fetch). To my knowledge the `json:true` would send the body as JSON, set that header, and parse the results as JSON.

For these routes that return file data streams, the move to `json:false` is correct, as is your change to the first one to send the body as JSON but not expect JSON back.

PR #1246, opened 2026-06-09
ACAS-988: Migrate acas base to node:20.20.2-slim; simplify launcher

brianbolt:
## Problem
The acas runtime base `quay.io/centos/centos:stream9` is a full distro on a rolling tag — violating the [container image security baselines](https://schrodinger.atlassian.net/wiki/spaces/LDE/pages/1940160531) (minimal, version-pinned runtime).

## Change
- **Base → `node:20.20.2-slim`** (minimal, version-pinned). `acas-oss` is a *base* image that downstream (`acas_custom_schrodinger`) and the dev docker-compose run `gulp build`/`gulp dev` on top of, so the build toolchain (gulp, coffeescript, git) is retained; `python3`/`pip` remain for the LiveDesign integration scripts the app shells out to at runtime. Only the unused `forever` process manager is dropped.
- **`bin/acas.sh` slimmed 523 → 63 lines.** It was a legacy SysV-init script managing node via `forever`, plus Apache/rapache and R, and assumed CentOS (`/etc/init.d/functions`). In a container the runtime is the process manager, so it is now a thin launcher: prepare config → wait for roo + run `PrepareModuleConfJSON` → `exec` the app as PID 1 (`gulp dev` for `dev`, else `node app.js`). This also removes the CentOS-specific init-functions dependency so it runs on the slim base.

## Testing
Built and run via docker-compose: acas serves the UI on :3000 (401 login) and the API on :3001; module config generation and dev watch both work.

acasclient tests pass
loading data in bbchem version tests + open in livedesign and curve images inline in LD.

> Note: a downstream follow-up is needed in `acas_custom_schrodinger/Dockerfile` (its `yum install` line is a CentOS-ism that breaks now that this base is Debian).

Part of ACAS-988.

## Runtime Python deps for ldclient (truststore / PEP 668)

acas installs the LiveDesign `ldclient` at runtime with `--no-deps` (`ServerAPI` `Bootstrap.coffee`), so ldclient's dependencies must already be present in the image. Moving the base to Debian surfaced two issues:

- **`truststore`** — `ldclient` 2026.3 made `truststore` a required dependency: SSL verification now uses the native OS trust store instead of extracting certs to temp files (see schrodinger/livedesign-client#623). No acas base ever shipped `truststore`, so the `--no-deps` startup install fails with `ModuleNotFoundError: No module named 'truststore'`. This is an **ldclient version bump, not a Debian-specific issue** — the old CentOS base hits it too against ldclient ≥ 2026.3. Fixed by pre-baking `truststore>=0.9.0` into the base.
- **PEP 668 (externally-managed environment)** — Debian marks its system Python as externally managed, so the runtime `pip install --user` of ldclient is refused. Fixed with `ENV PIP_BREAK_SYSTEM_PACKAGES=1` so runtime installs succeed without editing the acas pip commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,45 @@
-FROM quay.io/centos/centos:stream9
+# acas base image.
+#
+# This is a *base* image: downstream images (acas_custom_schrodinger) and the
+# dev docker-compose run `gulp build` / `gulp dev` on top of it, so the Node
+# build toolchain (gulp, coffeescript, git) must remain installed here. The
+# security-baseline win is the runtime base itself: the former full-distro,
+# rolling-tag quay.io/centos/centos:stream9 is replaced by the minimal,
+# version-pinned node:20.20.2-slim.
+FROM node:20.20.2-slim
 
-# Update
-RUN \
-  dnf update -y && \
-  dnf upgrade -y && \
-# tar for pulling down node
-# git required for some npm packages
-# postgresql for utility scripts
-  dnf install -y tar git && \
-  dnf install -y fontconfig urw-fonts iputils postgresql && \
-  dnf clean all
+# Runtime + build deps (Debian-slim equivalents of the old centos set):
+#  - git: required by some npm packages and downstream gulp builds
+#  - curl: bin/acas.sh polls the roo persistence service before starting
+#  - python3 + pip: the LiveDesign integration scripts the app shells out to
+#  - fontconfig + fonts-urw-base35: parity with the old fontconfig/urw-fonts
+RUN apt-get update && apt-get install -y --no-install-recommends \
+      git tar curl ca-certificates \
+      python3 python3-pip \
+      fontconfig fonts-urw-base35 \
+  && ln -sf /usr/bin/python3 /usr/bin/python \
+  && rm -rf /var/lib/apt/lists/*
 
-#Install python dependencies
-RUN   dnf install -y python311 python3.11-pip initscripts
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,40 +1,53 @@
-FROM quay.io/centos/centos:stream9
+# acas base image.
+#
+# This is a *base* image: downstream images (acas_custom_schrodinger) and the
```
bffrost: I find this comment a little bit misleading / confusing and it also has commentary specific to this one-time change.

The part that's misleading to me is highlighting it as a *base* image, which implies that it's not the runtime image. However currently the [acas_custom_schrodinger Dockerfile](https://github.com/schrodinger/acas_custom_schrodinger/blob/main/Dockerfile) uses this base image as the runtime image as well.

The "The security-baseline win is the runtime base itself: the former full-distro, rolling-tag quay.io/centos/centos:stream9 is replaced by the minimal, version-pinned node:20.20.2-slim." part of the comment can be removed since that's really just context on this PR, not something that needs to permanently be in this file.

bffrost (approved):
LGTM but I have some quibbles with one of the comments Claude added that I think are worth cleaning up before you merge.

Dockerfile
brianbolt: Agreed and thanks for commenting. I've pulled this whole comment out and made some other tweaks to the comments as they aren't relevant.

PR #1248, opened 2026-06-15
ACAS-923: Preserve protocol experiment 404 responses

bffrost:
## Description

The route for fetching experiments by protocol code was returning 500 error when protocol didn't exist. I papered over this with acasclient code to convert 500 errors into `None` results, which then obfuscated real 500 errors like those caused by Java OOM errors.

The `getFromACASServerInternal` function (used very widely in this codebase) converts any non-200 responses into 500s, so instead of changing that ubiquitous behavior, I've added a narrow alternative that preserves status codes on responses.

This PR is part 2 of 3:
- https://github.com/mcneilco/acas-roo-server/pull/518 fixes Roo error throwing to yield a 404
- This PR updates node to bypass "throw 500 on any non-200 response" rule built into node layer
- https://github.com/mcneilco/acasclient/pull/189 acasclient test & removal of workaround that was converting 500s to None

## How Has This Been Tested?

- On local docker compose with Roo & ACAS changes, hit http://localhost:3000/api/experiments/protocolCodename/FAKECODE and confirmed it returns 404 status and body `[]`

### DIFF
--- a/modules/ServerAPI/spec/serviceTests/ServerUtilityFunctionsMochaSpec.coffee
+++ b/modules/ServerAPI/spec/serviceTests/ServerUtilityFunctionsMochaSpec.coffee
@@ -18,6 +18,24 @@ parseResponse = (jsonStr) ->
 
 
 describe "Server Utiilty Function Tests", ->
+	describe "getFromACASServerInternalPreserveResponseStatus", ->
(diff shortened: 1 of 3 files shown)


brianbolt (approved):
LGTM

PR #1249, opened 2026-06-16
ACAS-994: Experiment retention policy (frontend field + folders-for-codes)

brianbolt:
## ACAS-994 — Experiment retention policy (frontend)

Frontend half of the experiment retention feature (backend: mcneilco/acas-roo-server#519).

### What it adds
- **Per-protocol retention field** on the protocol editor: "Deleted experiment retention (days)" (`numericValue / deleted experiment retention days`). Admins can edit it; other users see it read-only. Blank = keep indefinitely.
- **Read-only `POST /api/experiments/folders-for-codes`** route: resolves each experiment code to its folder path (relative to the data-files root) via `ControllerRedirectConf`. roo calls this during the purge and deletes the folders on its mounted volume.
- Documents the `server.experiment.retention.*` config (enabled [on by default], cron, cronZone, batchSize, globalDays [unset = no global policy]) with defaults in `conf/config.properties.example`; folder-path resolution reuses the existing `server.service.persistence.filePath` and `server.nodeapi.path`.

### Testing
Validated end-to-end against a local stack with the roo backend (#519); folder-path resolution exercised during real purges. See the *Manual end-to-end test procedure* comment on mcneilco/acas-roo-server#519.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/ServerAPI/conf/ExperimentConfJSON.coffee
+++ b/modules/ServerAPI/conf/ExperimentConfJSON.coffee
@@ -182,6 +182,12 @@
 			,
 				typeName: "codeValue"
 				kindName: "agonist batch code"
+			,
+				typeName: "dateValue"
+				kindName: "database deleted date"
+			,
+				typeName: "dateValue"
+				kindName: "files deleted date"
 			]
 
 		labeltypes:
--- a/modules/ServerAPI/conf/ExperimentConfJSON.coffee
+++ b/modules/ServerAPI/conf/ExperimentConfJSON.coffee
@@ -182,12 +182,6 @@
 			,
 				typeName: "codeValue"
 				kindName: "agonist batch code"
-			,
-				typeName: "dateValue"
-				kindName: "database deleted date"
-			,
-				typeName: "dateValue"
-				kindName: "files deleted date"
 			]
 
 		labeltypes:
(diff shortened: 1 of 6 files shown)


modules/ServerAPI/conf/ExperimentConfJSON.coffee
```diff
@@ -182,6 +182,12 @@
 			,
 				typeName: "codeValue"
 				kindName: "agonist batch code"
+			,
+				typeName: "dateValue"
+				kindName: "database deleted date"
```
bffrost: Are these experiment values used anywhere? They look like a possible holdover from the version of this idea where the top-level Experiment record was kept around, but the current roo implementation deletes the Experiment entries too.

brianbolt: Removed these from ExperimentConfJSON. They were leftovers from the earlier design where the top-level Experiment record would remain and could hold deletion markers. The current roo purge deletes the Experiment rows and only reads the protocol-level “deleted experiment retention days” value, so the experiment-level DB/files deleted date values are not used.

bffrost (approved):
LGTM, sorry for the delayed re-review

PR #1250, opened 2026-06-26
ACAS-1001 Fix auto dry-run report download

brianbolt:
## Summary

Fix the Standardization History `Dry Run Report` download route so callback-free GET streaming requests actually start and forward the SDF response back to the browser.

## Root Cause

`standardizationAutoDryRunReportFile` created a streaming request to the persistence endpoint and piped it to the Express response, but the callback-free GET stream was never ended. In this request adapter path, that meant the upstream request could remain idle, leaving the browser waiting on the download navigation.

## Changes

- Start the upstream GET stream explicitly with `upstream.end()`.
- Forward upstream `data` chunks and close the Express response on upstream `end`.
- Preserve existing header forwarding behavior for content type, disposition, and length.
- Add a regression spec proving the route starts the upstream stream and forwards the SDF response.

## Validation

- Ran focused Standardization route regression with a lightweight CoffeeScript/Mocha shim: `PASS starts the upstream report stream and forwards the SDF response`.
- Compiled `modules/Standardization/src/server/routes/StandardizationRoutes.coffee` with `coffee -c`.
- Compiled `modules/Standardization/spec/serviceTests/StandardizationRoutesSpec.coffee` with `coffee -c`.
- Hot-patched `qa-demo-26-3` using the existing `acas-custom-config-map` route mount, restarted `deployment/acas`, verified rollout completed, and confirmed the mounted route file contains `upstream.end()`.
- Manually confirmed via the UI that the download works with the patched file.

### DIFF
--- a/modules/Standardization/spec/serviceTests/StandardizationRoutesSpec.coffee
+++ b/modules/Standardization/spec/serviceTests/StandardizationRoutesSpec.coffee
@@ -1,0 +1,112 @@
+assert = require 'assert'
+EventEmitter = require('events').EventEmitter
+fs = require 'fs'
+Module = require 'module'
+path = require 'path'
+
+acasHome = '../../../..'
+
+describe "Standardization Routes", ->
+	originalLoad = null
+	routePath = null
+	fakeUpstream = null
+	requestedOptions = null
+
+	loadRoutesWithFakeRequest = ->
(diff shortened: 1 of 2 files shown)


bffrost (approved):
LGTM

PR #1251, opened 2026-06-26
ACAS-1000 show delete dependency errors

brianbolt:
## Summary
- Preserve CmpdReg admin delete error responses, including plain-text salt delete messages, instead of replacing them with a generic error.
- Propagate CodeTables delete status and parsed response bodies through the server route.
- Display parsed delete messages in the CodeTables and Salt delete modals instead of raw JSON.

## Root Cause
The CmpdReg admin delete proxy only forwarded dependency details for one response shape. Salt deletes return a plain-text dependency message, which caused JSON parsing to fail and the route to fall back to `{"error":"something went wrong :("}`. CodeTables deletes also dropped the downstream status and read the wrong response field from the request adapter.

## Validation
- Ran CoffeeScript compile checks for all five changed files with `coffee -c -p`.
- Hotfixed and manually verified on `qa-demo-26-3`: Vendors, Stereo Categories, and Salts now show dependency messages; Assay Scientist unused delete path succeeded; CmpdReg Scientist skipped because no entries existed in QA.
Screenshot 2026-06-26 at 2 36 07 PM

Screenshot 2026-06-26 at 2 36 15 PM
Screenshot 2026-06-26 at 3 09 54 PM

### DIFF
--- a/modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
+++ b/modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
@@ -149,8 +149,7 @@ exports.deleteCodeTableInternal = (codeTableEntry, callback) ->
 				console.log 'got ajax error trying to delete code table'
 				console.log error
 				console.log json
-				callback response.statusCode, response.json
+				callback response.statusCode, json
 		)
-
 
 
--- a/modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
+++ b/modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
@@ -95,7 +95,7 @@ exports.postCodeTableInternal = (codeTableEntry, callback) ->
 				console.log 'got ajax error trying to save new code table'
 				console.log error
 				console.log json
-				callback response.statusCode, response.json
+				callback response.statusCode, json
 		)
 
 exports.putCodeTable = (req, resp) ->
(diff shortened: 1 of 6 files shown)


modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
```diff
@@ -149,8 +149,7 @@ exports.deleteCodeTableInternal = (codeTableEntry, callback) ->
 				console.log 'got ajax error trying to delete code table'
 				console.log error
 				console.log json
-				callback response.statusCode, response.json
```
bffrost, line 152: I see a bunch more `response.json` statements in code above, so I'm not confident this point solution is the right one. Is there something systematically wrong with the request adapter?

bffrost (changes requested):
The fix to `CmpdRegAdminRoutes` to tolerate both plain string and JSON responses LGTM, as do the helpers to get a cleanly formatted error message.

I have low confidence on one of the fixes (`response.json` vs `json`) because that pattern shows up elsewhere. I'd like to see more about why that change was necessary

modules/Components/src/server/routes/CodeTableServiceRoutes.coffee
brianbolt: Thanks. From what I can tell these were always broken because response.json didn't work. I searched for remaining response.json usages. The only stale request-adapter callback usages were the CodeTableService error paths and the ProjectModule log line, and those are now fixed as part of this PR. The remaining usages are native fetch responses, where response.json() is a real function and is expected.

bffrost (approved):
LGTM, thanks for the explanation and the additional fixes

PR #1252, opened 2026-07-02
ACAS-999 support SAML metadata URL

brianbolt:
## Summary
- Add `server.security.saml.metadataUrl` as an optional SAML IdP metadata source.
- Fetch and parse IdP metadata to derive SAML `entryPoint`, `issuer`, and `idpCert` when the URL is configured.
- Preserve the existing explicit `entryPoint` / `issuer` / `cert` behavior when `metadataUrl` is empty.
- Add focused service tests for metadata parsing, explicit fallback, and fetch failure handling.

## Implementation notes
- `xml2js` was chosen as a small, common Node XML parser that works cleanly from the existing CoffeeScript server code.
- It gives structured access to SAML metadata attributes and nested certificate elements, while avoiding fragile regex/string parsing of XML.
- SAML metadata often uses XML namespaces, so the helper normalizes local tag names while extracting the fields ACAS needs.
- Tradeoff: `xml2js` is a generic XML parser, not a SAML-aware validation library. This PR only extracts the configured metadata fields; the IdP issuer versus SP issuer portability concern is called out as follow-up work.
- `startApp().catch(...)` was added because SAML strategy setup now awaits metadata fetch/parse. If metadata is unreachable or malformed, ACAS logs a clear startup error and exits non-zero.

## Notes
- Local Okta validation confirmed both the metadata URL path and the explicit settings path can authenticate when the ACAS callback host and SAML role sync are configured correctly.
- Follow-up work is being filed for the portability concern around metadata-derived IdP entity IDs versus `passport-saml` SP `issuer` semantics.

## Validation
- `docker compose exec -T acas sh -lc "./node_modules/.bin/gulp coffee:root coffee:serverAPI coffee:serviceTests copy:conf coffee:buildUtilities execute:prepare_config_files && ./node_modules/.bin/mocha --reporter spec src/spec/ServerAPI/serviceTests/SamlMetadataSpec.js --timeout 5000 --no-bail --check-leaks"`
- Result: `6 passing`

### DIFF
--- a/conf/config.properties.example
+++ b/conf/config.properties.example
@@ -359,6 +359,9 @@ client.security.authstrategy=${server.security.authstrategy}
 # Examples configs using okta but can use any SAML login IDP
 
 server.security.saml.use=false
+# When set, metadataUrl is used as the source of truth for entryPoint, issuer, and cert.
+# Leave metadataUrl empty to use the explicit entryPoint, issuer, and cert settings below.
+server.security.saml.metadataUrl=
 server.security.saml.entryPoint=https://<organization>.okta.com/app/<appName>/<issuerID>/sso/saml
 server.security.saml.issuer=http://www.okta.com/<issuerID>
 server.security.saml.audience=false
--- a/conf/config.properties.example
+++ b/conf/config.properties.example
@@ -359,14 +359,21 @@ client.security.authstrategy=${server.security.authstrategy}
 # Examples configs using okta but can use any SAML login IDP
 
 server.security.saml.use=false
-# When set, metadataUrl is used as the source of truth for entryPoint, issuer, and cert.
-# Leave metadataUrl empty to use the explicit entryPoint, issuer, and cert settings below.
+
+# When set, metadataUrl is used to derive the IdP entryPoint, issuer, and cert.
+# Leave metadataUrl empty to use the explicit IdP settings below.
 server.security.saml.metadataUrl=
+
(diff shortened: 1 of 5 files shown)


conf/config.properties.example
```diff
@@ -359,6 +359,9 @@ client.security.authstrategy=${server.security.authstrategy}
 # Examples configs using okta but can use any SAML login IDP
 
 server.security.saml.use=false
+# When set, metadataUrl is used as the source of truth for entryPoint, issuer, and cert.
```
bffrost: It also sets `audience`. While this comment is helpful, it'd be nice to reorder the SAML configs and add a few more comments. I'm thinking:
- metadataUrl the way you have it
- explicit configs (entryPoint, issuer, audience, cert, disableRequestedAuthnContext) that are alternatives to the metadataUrl
- ACAS-specific configs like the role sync, user attributes, and profileAttributeToSystemRoles

After this PR is merged we'll also need to update the docs on Confluence / Google Docs to make this new simpler config clear.

bffrost (changes requested):
Requested some doc changes in the config file.

Aside from this, which IdPs have you tested this against @brianbolt ?

conf/config.properties.example
bffrost: Thanks for the reorganization. I see I was wrong about `audience`

bffrost (approved):
LGTM

PR #1253, opened 2026-07-02
ACAS-1004: generic service-token login endpoint (POST /login/token)

brianbolt:
## Summary
Adds a **generic trusted-issuer service-token login** endpoint, `POST /login/token`, so a trusted external service can obtain an authenticated ACAS session on a user's behalf without a password or browser — the missing piece for headless/SSO integrations (Jira **ACAS-1004**).

## How it works
- Accepts `Authorization: Bearer <RS256 JWT>`.
- Verifies signature via JWKS (`jwks-rsa`), plus `iss` ∈ trustedIssuers, `aud`, and `exp` (`jsonwebtoken`).
- Maps the token's `user_email` to an author via the existing `csUtilities.getUser`, gates on `checkHasRole(user, loginRole)`, then `req.login` to establish the normal Passport session and returns the `connect.sid` (Set-Cookie + JSON body).
- Reuses ACAS's existing session, role, and per-project ACL machinery — the resulting session is bounded by that user's own permissions.

## Vendor-neutral by design
No external system is named in code or config. Trust is configured via generic values (`server.security.serviceToken.jwksUrl / trustedIssuers / audience`), so any conformant RS256/JWKS issuer can use it. This is delegated trust equivalent to the SAML IdP — see the config block next to `server.security.saml.*`.

## Config (feature-gated OFF by default)
```
server.security.serviceToken.use            = true
server.security.serviceToken.jwksUrl        = <trusted-issuer-jwks-url>
server.security.serviceToken.trustedIssuers = <trusted-issuer>
server.security.serviceToken.audience       = acas
server.security.serviceToken.clockSkewSec   = 30
```
Unconfigured instances behave exactly as today (`use=false` → endpoint returns 404).

## Behavior / fail-closed
- 404 when disabled; 401 on missing/invalid/expired/wrong-aud/untrusted-iss/bad-sig token or missing email; 403 when the author lacks a login role; 200 + `connect.sid` on success.

## Files
- `modules/Login/src/server/routes/ServiceTokenAuth.coffee` — RS256/JWKS verifier (new).
- `modules/Login/src/server/routes/loginRoutes.coffee` — `tokenLogin` handler + route in `setupRoutes` (web server only).
- `modules/Login/spec/serviceTokenAuthSpec.js` — verifier unit tests (valid / bad-sig / wrong-aud / wrong-iss / expired).
- `conf/config.properties.example`, `package.json` — config keys + `jsonwebtoken`/`jwks-rsa`.

## Testing
- Verifier unit tests pass in isolation (5/5).
- **Live end-to-end** on docker-compose ACAS: a valid RS256 token → `POST /login/token` **200 + connect.sid**; that cookie authenticated `GET /api/experiments/paginated` **200**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


modules/Login/src/server/routes/ServiceTokenAuth.coffee
```diff
@@ -0,0 +1,28 @@
...
+		clockTolerance: opts.clockSkewSec or 30
+	jwt.verify token, publicKeyPem, options, (err, claims) ->
+		return cb(err) if err
+		cb null, claims
+
+# Full path: read kid from the token header, fetch the signing key, verify.
+exports.authenticate = (token, config, cb) ->
+	decoded = jwt.decode token, complete: true
+	return cb(new Error('malformed token')) unless decoded?.header?.kid
+	client = jwksClient jwksUri: config.jwksUrl, cache: true, rateLimit: true
```
bffrost: Reconstructing this client object within this function defeats the purpose of the cache, which is important for avoiding repeated calls to the JWKS url on every call of the `tokenLogin` endpoint.

Can we store the `client` as a global variable so its in-memory cache of the fetched key can last longer?

bffrost (changes requested):
Overall this implementation looks good. It's great to see this using libraries from a reputable source (Auth0) and the pattern & configs look correct to me for verifying the JWT using the JWKS and looking up the user. I agree with having this disabled by default.

I noted one bug where the jwksClient won't be able to effectively cache because of the way it's instantiated. Once that's fixed, this LGTM.

modules/Login/src/server/routes/ServiceTokenAuth.coffee
brianbolt: Fixed in ec69c8de. Moved the client to a module-scope map keyed by JWKS URL (`_getClient`) so jwks-rsa's in-memory key cache persists across requests instead of being rebuilt each call. Verified in isolation: two `authenticate()` calls now trigger exactly one `GET /jwks.json` (was one per call before).

bffrost (approved):
LGTM

PR #1258, opened 2026-07-28
ACAS-1011: Sync SSO roles from LiveDesign

brianbolt:
## Summary

Adds an opt-in livedesign SAML role source. ACAS fetches mapped roles from LiveDesign at SSO login, synchronizes only the LiveDesign-managed ACAS/CmpdReg roles, and preserves the existing IdP flow by default.

The goal of this is to support LiveDesign admin interface owning the roles ACAS/CmpdReg users have.

## Configuration

- server.security.saml.roles.source=idp preserves existing SAML profile role synchronization.
- server.security.saml.roles.source=livedesign queries LiveDesign using the configured service account and uses the roles returned as the source to update the author roles instead of the IDP mapping.

## Validation

- Ran the new mocha tests

## Manual LiveDesign-role-sync verification completed:

 - With the user assigned only to LiveDesign AcasAdministrator, ACAS synchronized both ROLE_ACAS-ADMINS and the required ROLE_ACAS-USERS.
 - Changing the user to only AcasUser removed ROLE_ACAS-ADMINS and retained only ROLE_ACAS-USERS.
- With the user assigned only to CregAdministrator, ACAS synchronized both ROLE_CMPDREG-ADMINS and the required ROLE_CMPDREG-USERS.
 - CregUser was also verified to synchronize only its intended user role.
 - This confirms that administrator roles automatically imply their corresponding user roles when server.security.saml.roles.source=livedesign is configured. IdP-sourced role synchronization remains unchanged.


bffrost (approved):
LGTM.
The role name mapping seems stable enough to put into default configs, but I see you've kept the default behavior the same as before (idp-sourced roles).

PR #1261, opened 2026-08-14
ACAS-1018: remove unusable token-login session id

brianbolt:
## Summary

Remove `connect_sid` from the successful `POST /login/token` response body. The response continues to return `expires` and `user`.

## Root cause and security rationale

`req.sessionID` is the raw session-store identifier. Express-session sends the usable client credential as a signed `connect.sid` cookie (`s:<session-id>.<HMAC>`) in `Set-Cookie`; it verifies that HMAC before loading a session. The raw value in the JSON body therefore cannot authenticate a later request, while the signed cookie does.

The signed cookie is a bearer credential. Returning it in JSON would bypass the `HttpOnly` protection provided by `Set-Cookie` and make it more likely to be exposed to browser JavaScript or application logging. Cookie-capable clients should retain the `Set-Cookie` value with their normal cookie jar.

## Validation

- Added an in-process Express/Passport regression test that confirms the response body omits the raw session ID and the signed `Set-Cookie` cookie authenticates a subsequent request.
- `npx gulp coffee:routes coffee:serviceTests && npx mocha --reporter spec build/src/spec/Login/serviceTests/TokenLoginResponseSpec.js --timeout 5000 --check-leaks`
- No auth applications should read from the cookie, verified this works on a no auth application like Bunsen

Closes ACAS-1018


bffrost (approved):
LGTM, didn't realize this session ID was unusable without the cookie signing piece

